PR #21, opened 2022-11-28
make username and password optional

### LINKED ISSUE #18: Feature request: Make username,password optional parameter
If some software follows gNMI/gNOI spec to implement certain services ( eg: certificate management ), it would not be needed to add username based authentication to this service as authentication would be ensured in various other methods.

In such case, it becomes cumbersome to use "gnmic" as it always prompts for username and password. 

It would be nicer to make username/password prompt as optional or introduce a flag with which there should not be any need to username and password.

karimra:
Fix #18
This PR makes username/password optional for targets, gNMIc will not prompt the user for credentials when either username or password are not configured (via flags or config file)

### DIFF
--- a/config/targets.go
+++ b/config/targets.go
@@ -18,7 +18,6 @@ import (
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/openconfig/gnmic/types"
-	"golang.org/x/term"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


hellt:
if you agree with the doc updates done here https://github.com/openconfig/gnmic/commit/2c3ff06b1d2c54ed3ec3fb6fe25c0d827918bcd6 could you do `git cherry-pick  2c3ff06b1d2c54ed3ec3fb6fe25c0d827918bcd6` to merge those changes to the origin branch?

hellt (approved):
worth aligning the docs https://github.com/openconfig/gnmic/commit/2c3ff06b1d2c54ed3ec3fb6fe25c0d827918bcd6

PR #23, opened 2022-11-29
Add configurable timestamp precision to InfluxDB output

### LINKED ISSUE #17: Feature Request: configurable precision for export to InfluxDB
Current configuration assumes `ns` precision when writing to InfluxDB. However, if the "override ts" processor is used with the different precision, for example `s`, this will result with wrong timestamps in InfluxDB, because seconds are passed as nanoseconds.

The ideal would be having a configurable option for timestamp precision for export to InfluxDB and normalization of the event timestamp before passing it to the influxdb client library.

karimra:
fix #17

### DIFF
--- a/docs/user_guide/outputs/influxdb_output.md
+++ b/docs/user_guide/outputs/influxdb_output.md
@@ -31,6 +31,12 @@ outputs:
     # health check is disabled by default, can be enabled by setting the below field to any value other that zero.
     # with a minimum allowed period of 30s.
     health-check-period: 0s 
+    # defines the write timestamp precision, 
+    # one of `s` for second, `ms` for millisecond, `us` for microsecond and `ns` for nanoseconds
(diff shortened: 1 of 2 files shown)


dpajin:
Hi @karimra, I think the better approach would to set the precision option for the InfluxDB client using: https://github.com/influxdata/influxdb-client-go/blob/94dd2a00d884087ebdd5ec65692aeeaea1acca25/api/write/options.go#L93

In this way, influxdb client will use configured precision in the write calls to the DB:
```
http://<ip-address>:8086/api/v2/write?bucket=gnmic&org=&precision=ms
```
Otherwise, I think it will still use `ns`, just adjusting the timestamps to nanosec by filling zeros.

karimra:
Changed. Thanks, I wasn't aware fo that option.

hellt (approved):

Issue #14, opened 2022-11-25
Feature Request: using "/" in the subscription-name


dpajin:
It seems that it is not possible to use "/" in the subscription names. 
If I configure something like this:
```
subscriptions:  # container for subscriptions
  /system/state:
    paths:
       - "/system/state/hostname"
       - "/system/state/boot-time"
    stream-mode: sample
    sample-interval: 10s
```
this fails with the following error in the logs:
```
2022/11/25 15:23:37.160529 [gnmic] failed to subscribe: missing path(s) in subscription ''
```

karimra:
Yes this is a known behavior, the lib gnmic uses to parse and store the configuration uses / to separate nested levels.

dpajin:
OK, I have a workaround for what I was trying to achieve - having InfluxDB measurement names in the format "/system/state... " etc... 
Since subscription name is stored in the events into the `subscription-name` tag, I have used processors to change it (for example, replace "_" to "/"). However, InfluxDB output measurement is created from subscription name, before event goes though processors - so when the `subscription-name` is changed, the InfluxDB measurement name will still have original name.  

If this make sense for you, I would like to make a PR to make this configurable option in the InfluxDB output?
```
    # boolean, after the event is processed though the processors pipeline,
    # the final value of the 'subscription-name' tag will be used as the InfluxDB measurement name
    subscription-name-tag-as-measurement: false
```

karimra:
Agree on the idea, but I think there is no need for the configuration knob, it's fair to change the output behavior to use `subscription-name` as the measurement name by default. 
Right name it uses `ev.Name` instead of `ev.Tags["subscription-name]`. But it should default to the original name in case the tag subscription-name` does not exist (in case the user deletes it from the map of tags)

dpajin:
@karimra if the "subscription-name" is used for influxdb measurement name, do you think it should still be keept as a tag in the influxdb output, as it becomes a bit redundant?

karimra:
I think it's fine to delete it

dpajin-kentik:
@karimra should I close this issue, as we have a workaround merged with https://github.com/openconfig/gnmic/pull/24?

karimra:
if this is good enough for you, yes.

PR #24, opened 2022-11-29
use "subscription-name" tag as an influxdb measurement name

dpajin:
use "subscription-name" tag as an influxdb measurement name


dpajin:
closes #14

karimra (approved):
LGTM

PR #44, opened 2023-02-17
Add support of if-feature field for path generation

bclasse:
This PR brings support for the _if-feature_ field when paths are generated from a YANG module.

The if-feature statement is used to make conditional all instances of a YANG object subtree. In this PR, all the leafs present in a subtree with an if-feature statement, will be converted to a path with an extra parameter _FeatureList_ including the values of _if-feature_.

For example,

```  {
    "path": "/acl/cpm-filter/ipv6-filter/entry[sequence-id=*]/action/accept/log",
    "path-with-prefix": "/srl_nokia-acl:acl/cpm-filter/ipv6-filter/entry[sequence-id=*]/action/accept/log",
    "type": "boolean",
    "description": "When this is true, a log is created for each packet matching the entry\n\nFor IP packets matched by an IP filter entry the log entry contains the following information:\n- timestamp\n- filter name\n- filter entry sequence-id\n- incoming linecard\n- action: accept\n- IP protocol\n- packet-length\n- source IP address\n- source L4 port number (TCP/UDP packets)\n- destination IP address\n- destination L4 port number (TCP/UDP packets)\n- icmp-type (ICMP packets)\n- icmp-code (ICMP packets)",
    "default": "false",
    "namespace": "urn:srl_nokia/acl",
    "featurelist": [
      "srl_nokia-feat:jericho2 or srl_nokia-feat:jericho2cp"
    ]
  },
  {
    "path": "/acl/cpm-filter/ipv6-filter/entry[sequence-id=*]/action/accept/rate-limit/distributed-policer",
    "path-with-prefix": "/srl_nokia-acl:acl/cpm-filter/ipv6-filter/entry[sequence-id=*]/action/accept/rate-limit/distributed-policer",
    "type": "leafref",
    "description": "Reference to a policer",
    "namespace": "urn:srl_nokia/acl",
    "featurelist": [
      "not srl_nokia-feat:trident4",
      "not srl_nokia-feat:platform-7220-h2",
      "not srl_nokia-feat:platform-7220-h3",
      "not srl_nokia-feat:platform-7220-h4"
    ]
  },
  {
    "path": "/acl/cpm-filter/ipv6-filter/entry[sequence-id=*]/match/dscp-set",
    "path-with-prefix": "/srl_nokia-acl:acl/cpm-filter/ipv6-filter/entry[sequence-id=*]/match/dscp-set",
    "type": "dscp",
    "description": "A list of DSCP values to be matched for incoming packets. An OR match should be performed, such that a packet must match one of the values defined in this list. If the field is left empty then any DSCP value matches.",
    "namespace": "urn:srl_nokia/acl",
    "featurelist": [
      "srl_nokia-feat:ip-acl-dscp-set"
    ]
  },
  {
    "path": "/acl/cpm-filter/ipv6-filter/entry[sequence-id=*]/match/next-header",
    "path-with-prefix": "/srl_nokia-acl:acl/cpm-filter/ipv6-filter/entry[sequence-id=*]/match/next-header",
    "type": "ip-protocol-type",
    "description": "An IPv6 packet matches this condition if its first next-header field (in the IPv6 fixed header) contains the specified value",
    "namespace": "urn:srl_nokia/acl"
  },

### CODE REVIEW

File: app/path.go [hunk 1]
```diff
@@ -37,13 +37,14 @@ type pathGenOpts struct {
...
-	IsState        bool   `json:"is-state,omitempty"`
-	Namespace      string `json:"namespace,omitempty"`
+	Path           string   `json:"path,omitempty"`
+	PathWithPrefix string   `json:"path-with-prefix,omitempty"`
+	Type           string   `json:"type,omitempty"`
+	Description    string   `json:"description,omitempty"`
+	Default        string   `json:"default,omitempty"`
+	IsState        bool     `json:"is-state,omitempty"`
+	Namespace      string   `json:"namespace,omitempty"`
+	FeatureList    []string `json:"featurelist,omitempty"`
```
hellt, line 47: imho it would be nicer to name it `if-features` to have a clear mapping to the yang naming

app/path.go
```diff
@@ -228,18 +231,49 @@ func collectSchemaNodes(e *yang.Entry, leafOnly bool) []*yang.Entry {
...
+				Kind:        e.Kind,
+				Config:      e.Config,
+				Prefix:      e.Prefix,
+				Mandatory:   e.Mandatory,
+				Dir:         e.Dir,
+				Key:         e.Key,
+				Type:        e.Type,
+				Exts:        e.Exts,
+				ListAttr:    e.ListAttr,
+				Extra:       make(map[string][]interface{}),
```
hellt, line 256: given `any` is introduced in go1.18 for an empty interface it might be good to introduce new code with it

app/path.go
```diff
@@ -228,18 +231,49 @@ func collectSchemaNodes(e *yang.Entry, leafOnly bool) []*yang.Entry {
...
+			collected = append(collected, f)
 		case e.ListAttr != nil: // list
 			fallthrough
 		default: // container
 			if !leafOnly {
 				collected = append(collected, e)
 			}
+			if e.Extra["if-feature"] != nil {
+				for _, myleaf := range collected {
+					if myleaf.Extra["if-feature"] != nil && myleaf.Extra["if-feature"][0] != nil {
```
karimra, line 270: Shouldn't this check if `len(e.Extra["if-feature"]) != 0` ?
Also, if `myleaf.Extra["if-feature"] == nil`, initialize it.

app/path.go
```diff
@@ -228,18 +231,49 @@ func collectSchemaNodes(e *yang.Entry, leafOnly bool) []*yang.Entry {
...
+				Extra:       make(map[string][]interface{}),
+			}
+			for k, v := range e.Extra {
+				f.Extra[k] = v
+			}
+			collected = append(collected, f)
 		case e.ListAttr != nil: // list
 			fallthrough
 		default: // container
 			if !leafOnly {
```
bclasse, line 265: The leafOnly case isn't clean with this PR : when running gnmic path with the --with-non-leaves, the conditions are not associated properly with each container path (in short, there is duplication in the FeatureList). Has to be corrected, or if-feature could be added as an optional flag, mutually exclusive with --with-non-leaves. What do you think @hellt @karimra ?

app/path.go
bclasse: Noted for `len`
If `myleaf.Extra["if-feature"] == nil`, it is defined in the else block, line 273; is it enough ?

app/path.go
karimra: Can we do it as we discussed on the call? Finish the collection then resolve the features recursively?

karimra (approved):

karimra:
@bclasse thanks for the contribution

Issue #36, opened 2023-01-25
Possible to suppress empty messages?


bewing:
I am dealing with a system that is mapping an internal database to arbitrary path elements in gNMI where keys are part of the elem path string, instead of actually being filterable keys.  As a result, I am doing a LOT of filtering on events, and in some cases, end up with messages (lists of Events) that are comprised of just empty events after using the `event-drop` filter to delete unwanted data.  The `file` output looks like this:

```
[
  {}
]
[
  {},
  {}
]
[
  {}
]
[
  {},
  {}
]
[
  {},
  {}
]
[
  {},
  {}
]
[
  {}
]
[
  {},
  {}
]
```

Is it possible to completely remove empty events from the message, and drop/suppress empty messages instead of writing them to outputs further downstream?

karimra:
I think it should be doable to add a knob under the output, something like `ignore-empty: true`

Sparc0:
would like this setting added also.

karimra:
@bewing , @Sparc0  #61 makes the drop processor remove matching messages from the events list. The messages won't make it to the downstream outputs.

PR #61, opened 2023-03-06
drop messages instead of setting them to empty struct

karimra:
#36

### DIFF
--- a/formatters/event_drop/event_drop.go
+++ b/formatters/event_drop/event_drop.go
@@ -116,7 +116,8 @@ func (d *Drop) Init(cfg interface{}, opts ...formatters.Option) error {
 }
 
 func (d *Drop) Apply(es ...*formatters.EventMsg) []*formatters.EventMsg {
-	for _, e := range es {
+	toDrop := make([]int, 0, len(es))
+	for i, e := range es {
 		if e == nil {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


formatters/event_drop/event_drop.go
```diff
@@ -180,3 +185,15 @@ func (d *Drop) WithLogger(l *log.Logger) {
...
+	lastDropIndexes := len(dropIndexes) - 1
+	for i, dropIndex := range dropIndexes {
+		if i == lastDropIndexes {
+			es = es[:dropIndex]
+			continue
+		}
+		copy(es[dropIndex:], es[dropIndexes[i+1]:])
+	}
+	return es
+}
```
hellt, line 199: @karimra is the idea of this function to drop elements from `es` whos indexes match the values in `dropIndexes`?

e.g.
```go
dropIndexes := []int{0, 2, 3}
es := []string{"A", "B", "C", "D", "E", "F"}
// should return B E F ?
```

karimra: Yes but this doesn't seem to work properly when the first message is to be dropped, will rework it.

hellt: yep, I personally find this solution a very easy to read

https://goplay.tools/snippet/fcH6_GTddi1

it has a O(n) complexity, but it seems that the order of the events needs to be preserved, thus the re-slicing (shifting) is not considerably faster on the number of events

karimra: My point was to not have to allocate a new event slice to append the allowed messages.

hellt (approved):

Issue #76, opened 2023-03-20
Be able to set Kafka event key


Sparc0:
I would like a config parameter so that gnmic does not set the kafka key to null. 
Either the config parameter can be boolean and you use source + subscription_name as event key or if i could specifiy what fields to use as key.

I want this so that i can [guarantee message ordering](https://medium.com/@felipedutratine/kafka-ordering-guarantees-99320db8f87f) if i have a kafka topic with more then 1 partition.
The issue i get today with remote_write towards cortex is that i get [err: out of order sample ](https://www.robustperception.io/debugging-out-of-order-samples/)

PR #77, opened 2023-03-21
add `insert-key` attribute under kafka output config

karimra:
#76


Sparc0:
I built this pull and it did not work as i expected. 
Not sure if you are done or not but when the message is produced into the topic they key need to be set there. 

```
msg := &sarama.ProducerMessage{
	Topic: k.Cfg.Topic,
	Key: <dynamic value based on message>
	Value: sarama.ByteEncoder(b),
}
```

Sparc0:
Ah read the code wrong i had InsertKey: true in my config. 
Works as expected thanks you.

PR #89, opened 2023-04-03
Target subscribe state metric

mc-dn:
feature: add api server gauge metrics for target subscribe state

What I'm trying to accomplish here is to expose the subscription state of loaded targets for reporting or alerting.  If this isn't the right way to implement such a feature, please point me in the right direction and I'll implement it in a way that better follows the architecture of gNMIc.


karimra:
Thanks for the PR, I read through it briefly. 
I didn't see a way for the gauge to be set back to 0 when the target is not subscribed to anymore.
Maybe a `defer subscribeTargetState.WithLabelValues(t.Config.Name).Set(0)` at the start the target goroutine could do the job ?

mc-dn:
Still haven't forgotten about this.  Something with `defer()` coming soon.

mc-dn:
This GitHub account will be closed.  I may work on this through another account.

PR #140, opened 2023-06-02
change sftp known_hosts handling

steiler:
@karimra I've discovered a flaw regarding the readSFTPFile() function.
It makes use of the getHostKey() method. This will go through the known_hosts file and try to find the entry for the host to connect to. It does so while assuming the actual IP is the first field in the KH file.
However meanwhile KH files are usualy(?) hashed.
Resulting in lines like this:
```
|1|wlPQdgFoYgYsqG6ae20lYopRLPI=|p61txQKmb+Hn49dsD+v0CNuEKd4= ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAzhZmG33G/3FG3vm0eDdyX1u++i0ceakIkJNgDxVVy6MpodRrpwqXXQj8/OGT
Iwb4YpRXGuL3236IkGugI9GUgFd00UNjMSMt3pqob4hKsEzADl7YfZeV1X7X0b617nze0otdO7TwDMlQ/5KWUwdUoxg50VfpieTzcOpUN/G4J159iKZ41iSF7o4vI+fYisX8y5rJ1BRbt1HO0Gi7w9HZ8tN0B
0glM6JKyoE8TjvbZAeD9PWIWp9JpG1KTY4yXTV1B1CyvtxjRqTMm8mcb+gSGGvv6mSlWCNxJnlXhp91F2GtmgzKsE3FjcMUfkn3c0+P0bKaR8L3GtbyaXJmtDX4xQ==
```
Hence the `hostKey` var will always be empty, resulting in utilizing the `ssh.InsecureIgnoreHostKey()` ssh.HostKeyCallback.

This PR implements a custom `HostKeyCallback`. That will try to utilize the `known_hosts` file provided. If this is not successfull it will basically behave like `InsecureIgnoreHostKey()` but in addition log a warning, to inform the user about the potential security issue.

Now, I've noticed gnmic is not(?) utilizing logging as intended in this PR. So the question is, if this is usefull or should it be further simplified, solely relying on `InsecureIgnoreHostKey()` becasue thats in the end what happens with the actual implementation.

### DIFF
--- a/utils/file.go
+++ b/utils/file.go
@@ -22,8 +22,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	log "github.com/sirupsen/logrus"
-
 	"github.com/jlaffaye/ftp"
 	"github.com/pkg/sftp"
 	"golang.org/x/crypto/ssh"
@@ -50,7 +48,7 @@ func ReadFile(ctx context.Context, path string) ([]byte, error) {
 	case strings.HasPrefix(path, "ftp://"):
 		return readFTPFile(ctx, path)
 	case strings.HasPrefix(path, "sftp://"):
-		return readSFTPFile(ctx, path)
+		return readSFTPFile(ctx, path, false)
 	default:
 		return readLocalFile(ctx, path)
 	}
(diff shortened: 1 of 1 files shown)


steiler:
Alright, so we'll simply rely on the insecure and the KnownHosts implementation. Configurable via a flag.
The flag will be exposed later on to the user by @karimra.

karimra (approved):

karimra:
Thanks for the contribution

PR #150, opened 2023-06-17
fix consul loader behaviour with multiple services

### LINKED ISSUE #147: Registering a target from a second consul service group deletes all other targets
I have two consul loader service groups setup with different subscriptions and configuration. If I register a device in the second group, gnmic deletes all devices from the first group and only monitors the new device.

```
loader:
  type: consul
  datacenter: lon
  address: 10.249.0.250:8500
  enable-metrics: true
  services:
    - name: junos-targets
      subscriptions:
        - interfaces
        - junos-bgp
    - name: cumulus-targets
      config:
        encoding: JSON
        skip-verify: true
      subscriptions:
        - interfaces
 ```

I have 30+ devices in the `junos-targets` group working correctly. If I add a device to `cumulus-targets` the logs show all the junos-targets being deleted despite them still being active in Consul. If I restart gnmic, it registers devices from both services.

```
10.249.0.250 - - [15/Jun/2023:15:48:22 +0000] "DELETE /api/v1/config/targets/router1 HTTP/1.1" 404 57
10.249.0.250 - - [15/Jun/2023:15:48:22 +0000] "DELETE /api/v1/config/targets/router2 HTTP/1.1" 404 59
10.249.0.250 - - [15/Jun/2023:15:48:22 +0000] "DELETE /api/v1/config/targets/router3 HTTP/1.1" 404 59
10.249.0.250 - - [15/Jun/2023:15:48:22 +0000] "DELETE /api/v1/config/targets/router4 HTTP/1.1" 404 51
10.249.0.250 - - [15/Jun/2023:15:48:22 +0000] "DELETE /api/v1/config/targets/router5 HTTP/1.1" 404 59
10.249.0.250 - - [15/Jun/2023:15:48:22 +0000] "DELETE /api/v1/config/targets/router6 HTTP/1.1" 404 58
2023/06/15 15:48:22.735425 [gnmic] adding target {"name":"cumulus-test01","address":"10.12.128.55:9339","username":"","password":"****","timeout":10000000000,"insecure":true,"skip-verify":true,"buffer-size":100,"retry-timer":10000000000,"log-tls-secret":false,"gzip":false,"token":""}
```

### PR DESCRIPTION
**Pull request by karimra**
fix #147

### DIFF
--- a/loaders/consul_loader/consul_loader.go
+++ b/loaders/consul_loader/consul_loader.go
@@ -45,7 +45,7 @@ func init() {
 		return &consulLoader{
 			cfg:         &cfg{},
 			m:           new(sync.Mutex),
-			lastTargets: make(map[string]*types.TargetConfig),
+			lastTargets: make(map[string]map[string]*types.TargetConfig),
 			logger:      log.New(io.Discard, loggingPrefix, utils.DefaultLoggingFlags),
 		}
 	})
@@ -54,9 +54,10 @@ func init() {
 type consulLoader struct {
 	cfg *cfg
 	// decoder        *consulstructure.Decoder
-	client         *api.Client
-	m              *sync.Mutex
-	lastTargets    map[string]*types.TargetConfig
+	client *api.Client
+	m      *sync.Mutex
+	// map of targets per service
+	lastTargets    map[string]map[string]*types.TargetConfig
 	targetConfigFn func(*types.TargetConfig) error
 	logger         *log.Logger
 	//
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: loaders/consul_loader/consul_loader.go [hunk 1]
```diff
@@ -54,9 +54,10 @@ func init() {
 type consulLoader struct {
 	cfg *cfg
 	// decoder        *consulstructure.Decoder
-	client         *api.Client
-	m              *sync.Mutex
-	lastTargets    map[string]*types.TargetConfig
+	client *api.Client
+	m      *sync.Mutex
+	// map of targets per service
+	lastTargets    map[string]map[string]*types.TargetConfig
```
hellt, line 60: NIT: maybe it makes sense to have a struct as a key for targets per service, so that it becomes less hairy when adding items to a nested map?

```
type TargetKey struct {
    Service, Target string
}
lastTargets := make(map[TargetKey]*types.TargetConfig)
```

hellt (approved):

loaders/consul_loader/consul_loader.go
karimra: In some functions I need to pass all the targets per service (for e.g loaders.Diff()), having nested maps allows to easily get that map without having to iterate over the lastTargets map and build a new one.

PR #195, opened 2023-08-02
make down and dirty build script to build this as an rpm

kajtzu:


kajtzu:
oops this wasn't supposed to come here but stay in my own repo ;)

PR #262, opened 2023-10-24
Add rate-limit processor

netixx:
Create a new event processor that rates-limits the amount of messages that arrive through a subscription.

This is useful to protect infrastructure behind the collector in case from high network change rate (due to eg. a network outage).


docs/user_guide/event_processors/event_rate_limit.md
```diff
@@ -0,0 +1,28 @@
+The `event-rate-limit` processor rate-limits each event with matching tags to the configured amount per-seconds.
+
+All the tags for each event is hashed, and if the hash matches a previously seen event, then the timestamp 
+of the event itself is compared to assess if the configured limit has be exceeded.
```
karimra, line 4: Minor typo: `limit has be exceeded.` --> `limit has been exceeded`

formatters/event_rate_limit/event_rate_limit.go
```diff
@@ -0,0 +1,135 @@
...
+
+	o.eventIndex, err = lru.New[string, int64](o.CacheSize)
+	if err != nil {
+		return fmt.Errorf("failed to initialize cache: %w", err)
+	}
+	return nil
+}
+
+func (o *RateLimit) Apply(es ...*formatters.EventMsg) []*formatters.EventMsg {
+	validEs := []*formatters.EventMsg{}
```
karimra, line 85: maybe worth allocating the capacity of the slice here: `validEs := make([]*formatters.EventMsg, 0, len(es))`

formatters/event_rate_limit/event_rate_limit.go
```diff
@@ -0,0 +1,135 @@
...
+		tagKeys[i] = tagKey
+		i++
+	}
+	sort.Strings(tagKeys)
+
+	for _, tagKey := range tagKeys {
+		h.Write([]byte(tagKey))
+		h.Write([]byte("="))
+		h.Write([]byte(e.Tags[tagKey]))
+		h.Write([]byte("\n"))
```
karimra, line 119: a minor optimisation here: 
Declare `[]byte("=")` and `[]byte("\n")` as package level variables instead of creating new ones for each event/tag

karimra (changes requested):

formatters/event_rate_limit/event_rate_limit.go
netixx: ok, it's not an issue if there are nil values in the slice that is returned by the processor then ?
I can have a counter inside the filter loop, that increments and pushes accepted events in the slice, but that means that there will be nil values at the end of the slice.

Or should I re-slice it before returning it ?

formatters/event_rate_limit/event_rate_limit.go
netixx: fixed

docs/user_guide/event_processors/event_rate_limit.md
netixx: done

formatters/event_rate_limit/event_rate_limit.go
netixx: I went for allocating the whole slice, and then slicing it at the end, based on the accepted count.
I actually never looked at make with more than one int :), but I can revert to that if you prefer.

karimra: When creating a slice you can specify both its length and capacity: `make([]any, len, cap)`. 
The capacity is defined as `the maximum number of elements that the underlying array can hold without reallocation.`
This allows appending elements to the slice up to the capacity without having to reallocate the underlying array. There will be no nil elements at the end of the array.
I think it's better to set the length to 0, the capacity to `len(es)` and use `append()`.

netixx: done :)

karimra (approved):

karimra:
Thanks for the PR, writing docs and making the requested changes. Looking forward to the next ones :)

PR #268, opened 2023-10-30
Fix tls-ca key for yaml

netixx:
There was a typo (I think) with yaml decoding key for tls-ca on tagets. It was "tlsca"

This aligns it with the other keys to be "tls-ca".


karimra (approved):

karimra:
Thanks for catching this.
For the record, that struct tag does not influence config reading. The YAML unmarshaling is based on the `mapstructure` struct tag.

netixx:
I was having the issue using "tls-ca" in my yaml config which wasn't taken into account (checked using the debug logs of the loader), so I think `mapstructure` is not used here ?

I see in the code that 
```
// unmarshal the bytes into a map of targetConfigs
	err = yaml.Unmarshal(b, result)
```
is done, so I do believe that the yaml tag is used ?

karimra:
Ah yes you are right, the `yaml` struct tags will be used with loaders. The `mapstructure` tag is used when reading the main config.

PR #274, opened 2023-11-02
Implement WriteEvent on file output so that it can be used by 'inputs' components

netixx:
WriteEvent is used by the "inputs" components to output their data.

Without this implementation,
things like
```
inputs:
  nats-replay:
    type: nats
    address: nats-server:4222
    subject: replay
    debug: true
    buffer-size: 1
    outputs:
      - file-dump

outputs:
  file-dump:
    type: file
    format: event
    debug: true
    split-events: true
    filename: /output.json
    concurrency-limit: 1
```

wont actually output anything to the file.

### DIFF
--- a/outputs/file/file_output.go
+++ b/outputs/file/file_output.go
@@ -267,7 +267,60 @@ func (f *File) Write(ctx context.Context, rsp proto.Message, meta outputs.Meta) 
 	}
 }
 
-func (f *File) WriteEvent(ctx context.Context, ev *formatters.EventMsg) {}
+func (f *File) WriteEvent(ctx context.Context, ev *formatters.EventMsg) {
+	select {
+	case <-ctx.Done():
+		return
+	default:
+	}
+	var evs = []*formatters.EventMsg{ev}
+	for _, proc := range f.evps {
+		evs = proc.Apply(evs...)
+	}
+	toWrite := []byte{}
+	if f.Cfg.SplitEvents {
+		for _, pev := range evs {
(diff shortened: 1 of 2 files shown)


outputs/file/file_output.go
```diff
@@ -267,7 +267,60 @@ func (f *File) Write(ctx context.Context, rsp proto.Message, meta outputs.Meta)
...
+			b, err = json.Marshal(evs)
+		}
+		if err != nil {
+			fmt.Printf("failed to WriteEvent: %v", err)
+			numberOfFailWriteMsgs.WithLabelValues(f.file.Name(), "marshal_error").Inc()
+			return
+		}
+		toWrite = append(toWrite, b...)
+		toWrite = append(toWrite, []byte(f.Cfg.Separator)...)
+	}
```
karimra, line 313: This part can be replaced with the function `outputs.Marshal()`. Just like in the `Write()` method [here](https://github.com/openconfig/gnmic/blob/main/outputs/file/file_output.go#L230)

karimra (changes requested):

outputs/file/file_output.go
netixx: ok, I saw that, but it requires a proto.Message...

I don't know how I could transform the `ev *formatters.EventMsg` to the correct proto.Message.
Could you give a pointer on how to do that ?

karimra (approved):

netixx:
Thanks!

Issue #259, opened 2023-10-19
Prometheus scrape based metrics drop frequently


spolack:
Hello,

first of all thank you for that amazing tool! Basically unlocks unlimited ways to mangle our metrics!

I'm still evaluating, but seem to have found the first bug. The metrics in the scrape endpoint are frequently dropping, feels like a timeout, but i can confirm from the API endpoint and debug log, that metrics are still received on the sample-interval from the subscription.

Config
```
subscriptions:
  cpu:
    paths:
      - "/components/component/cpu/utilization/state/instant"
    stream-mode: sample
    sample-interval: 15s

outputs:
  prom-output:
    type: prometheus
    listen: :8080
    path: /metrics 
    expiration: 60s 
    metric-prefix: "arista" 
    append-subscription-name: true
    override-timestamps: false
    export-timestamps: false
    strings-as-labels: false
    timeout: 10s
    debug: true
    target-template:
    num-workers: 1

api-server:
  address: :8081
  timeout: 10s
  enable-metrics: true
```

Metrics:
```
curl http://mgmt-r1-lab2-de:8080/metrics  
# HELP arista_cpu_components_component_cpu_utilization_state_instant gNMIc generated metric
# TYPE arista_cpu_components_component_cpu_utilization_state_instant untyped
arista_cpu_components_component_cpu_utilization_state_instant{component_name="CPU0",source="127.0.0.1:6030",subscription_name="cpu"} 1
arista_cpu_components_component_cpu_utilization_state_instant{component_name="CPU1",source="127.0.0.1:6030",subscription_name="cpu"} 1
arista_cpu_components_component_cpu_utilization_state_instant{component_name="CPU10",source="127.0.0.1:6030",subscription_name="cpu"} 4
arista_cpu_components_component_cpu_utilization_state_instant{component_name="CPU11",source="127.0.0.1:6030",subscription_name="cpu"} 4
arista_cpu_components_component_cpu_utilization_state_instant{component_name="CPU12",source="127.0.0.1:6030",subscription_name="cpu"} 2
arista_cpu_components_component_cpu_utilization_state_instant{component_name="CPU13",source="127.0.0.1:6030",subscription_name="cpu"} 1
arista_cpu_components_component_cpu_utilization_state_instant{component_name="CPU14",source="127.0.0.1:6030",subscription_name="cpu"} 8
[...]

Issue #169, opened 2023-07-04
Prometheus graphs traffic drops regularly


jasonyates-thg:
I'm using gnmic to monitor various Juniper MX devices, i'm subscribing to interface metrics and the sample interval is configured as 10s. My output is prometheus which is configured to scrape every 20s.

Once I graph my data, there is a routine drop in traffic across every device and every interface roughly every 60 minutes. At this point i'm unsure if it's something the Juniper device is doing, or something gnmic is doing or where to begin troubleshooting.

Example graphs from 2 different devices, different hardware (MX10k vs MX204) running different Junos software versions.

<img width="816" alt="Screenshot 2023-07-04 at 22 45 55" src="https://github.com/openconfig/gnmic/assets/125972762/71e4281c-014c-4710-a5f2-b3a60c609123">
<img width="813" alt="Screenshot 2023-07-04 at 22 46 13" src="https://github.com/openconfig/gnmic/assets/125972762/4498928d-fe08-4723-af94-1aaf53507b58">

karimra:
I would be surprised if this is coming from the routers if it's happening on all devices.
It looks like the period is more like 50minutes.

Could you share the prometheus output config used for gNMIc ? Do you have `export-timestamps` set to true there ?

I would also look at the promQL expression used in Grafana, I assume it uses the `rate()` function. 
Which interval is used for the rate function? It's usually recommended to use `$__rate_interval`.

jasonyates-thg:
I'd agree they are every 50 minutes. My output config is largely the default

```
# Configure outputs
outputs:
  prometheus:
    type: prometheus
    listen: :9804
    path: /metrics
    event-processors:
      - interface-descriptions
    service-registration:
      address: 10.249.0.250:8500
      service-address: 10.249.1.215
```

My promQL expression is using `irate` vs `rate`. If I change to `rate` the dips in the graph become less severe but they are still there. The expression is using` $__rate_interval`

Example query
`irate(interfaces_interface_state_counters_in_octets{source=~"router1|router2", description!=""}[$__rate_interval])*8`

irate:
<img width="2011" alt="Screenshot 2023-07-05 at 10 21 02" src="https://github.com/openconfig/gnmic/assets/125972762/ad594d73-5964-4864-8ea9-053de2dab841">

rate:
[...]

PR #277, opened 2023-11-05
Fix bug corrupting Prometheus timeseries __name__ label

alibresco:
I have also been experiencing issues like the ones described in #259 and #169 where Prometheus outputs are sometimes dropped or incorrect.

I believe this is due to a bug with how the `__name__` label gets added to the list of labels when constructing a timeseries. The `__name__` label is appended without a copy or assignment, meaning it will overwrite whatever was appended previously, so long as there is space remaining in the array. If GetLabels returns a slice with length == capacity, the bug does not seem to occur because append forces a copy.

A consequence of this bug is that a value can get written to the wrong timeseries.

I've written a test that reproduces this behavior. The test fails without the fix and passes with the fix.

I don't write a lot of Go, so let me know if I've gotten anything wrong here or there's a better way to solve this. I stumbled upon this issue while investigating incorrect gNMIC Prometheus writes that I was observing.

### DIFF
--- a/outputs/prometheus_output/prometheus_common.go
+++ b/outputs/prometheus_output/prometheus_common.go
@@ -143,14 +143,17 @@ func (m *MetricBuilder) TimeSeriesFromEvent(ev *formatters.EventMsg) []*NamedTim
 			fv = 1.0
 		}
 		tsName := m.MetricName(ev.Name, k)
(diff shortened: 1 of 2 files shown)


karimra:
@alibresco  This makes sense, I will try it out and merge if all good.

karimra (approved):

karimra:
Thanks for the contribution and for writing tests.

PR #285, opened 2023-11-10
Make the output/input fail when processor fails

netixx:
In the case where the event processor fails to initialise (could be for example a bad processor config, file not found, starlark syntax error), the previous behaviour was to ignore the processor.

However an event-processor can be crucial to the event processing and ignoring it could lead to garbage in the downstream metrics platform. It could lead for example to cardinality explosion.

This changes makes the input or output component fail if any processor in the chain fails to initilise.

### DIFF
--- a/pkg/app/inputs.go
+++ b/pkg/app/inputs.go
@@ -29,7 +29,12 @@ func (a *App) InitInput(ctx context.Context, name string, tcs map[string]*types.
 				go func() {
 					err := in.Start(ctx, name, cfg,
 						inputs.WithLogger(a.Logger),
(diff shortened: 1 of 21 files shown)


karimra (changes requested):
Maybe it would make sense to merge both `MakeEventProcessors` for inputs and outputs into a single function and put it under the `formatters` package ?
I know the inputs one does not include `actions` config for now. It can be left `nil` for inputs.

karimra (approved):

karimra:
Ended up making the changes myself, thanks for the contribution!

netixx:
Oh sorry I just saw you message right now!

thanks :)

Issue #272, opened 2023-11-01
Consul Loader ignoring tags


lucabrasi83:
Hi!

I'm using gnmic v0.33 and have setup the Consul loader as below with multiple entries using the same name but different tags in order to apply different subscriptions:

 

```yaml

loader:
  type: consul
  address: consul:8500
  datacenter: dc1
  enable-metrics: true
  services:
    - name: networks-gnmic-targets
      tags:
        - "vendor=arista"
        - "platform=eos"
      config:
        subscriptions:
          - netgenie_telemetry_proto
        outputs:
          - prom-output-proto
        insecure: true
    - name: networks-gnmic-targets
      tags:
        - "vendor=cisco"
        - "platform=iosxe"
      config:
        subscriptions:
          - netgenie_telemetry_json
        outputs:
          - prom-output-cisco-xe
        insecure: true

```

The tags are correctly assigned to the target service ID's in Consul:

 

```json

{

  "ID": "test-cat9k",
  "Service": "networks-gnmic-targets",
  "Tags": [
    "vendor=cisco",
    "platform=iosxe"

  ],

  "Meta": {},
  "Port": 50052,
  "Address": "10.0.46.11",
  "Weights": {
    "Passing": 1,
    "Warning": 1
  },

  "EnableTagOverride": false,
  "ContentHash": "fe321e05900eb3a8"

}

```

However, this seems to be ignored by gnmic as it just applies the config from the first service name:

 

`2023/11/01 23:20:58.784037 [gnmic] adding target {"name":"test-cat9k","address":"[10.0.46.11:50052](http://10.0.46.11:50052/)","username":"****","password":"****","timeout":10000000000,"insecure":true,"skip-verify":false,"subscriptions":["netgenie_telemetry_proto"],"outputs":["prom-output-proto"],"buffer-size":100,"retry-timer":10000000000,"log-tls-secret":false,"gzip":true,"token":""}`

 

Is it the correct way I'm configuring it and it's supposed to work ?

karimra:
gNMIc applies the config from the first service with a matching name. This was not designed to support services with the same name.
It should work if you create 2 services called:`networks-gnmic-targets-cisco` and `networks-gnmic-targets-arista`.

If you really need to create just one service I will need to write a PR to match against both the service name and the tags.

lucabrasi83:
Thanks @karimra using different services indeed work. So I guess I was getting confused by https://github.com/openconfig/gnmic/issues/106#issuecomment-1512021956

[...]

PR #296, opened 2023-11-14
Fix Consul Loader panic from #272

lucabrasi83:
As discussed in #272 this PR is regarding a fix to avoid returning nil when no matching service(s) are found in the Consul loader.

The `continue` statement is also effective when the configured service is not found in Consul.


karimra (approved):
Thanks for catching this. Looks like I broke it in the last release.

PR #307, opened 2023-12-01
POC, event processors using go-plugin

fredsod:
Hi,

I needed a way to dynamically add tags to Prometheus metrics and here is a POC using hashicorp go-plugin to build event processor plugins. In my case I would like to add metadata from our inventory to the metrics, but there could be many other use cases. 

gNMIc loads plugins from './plugins/' and the name of the binary is also the plugin name, it then communicates with the plugin over RPC. Each plugin needs to implement the EventProcessor interface, due to the RPC the logging setup is a bit different in the plugin. 

How to test the POC: Compile the examples/plugin/main.go to ./plugins/event-add-device_function and add 'event-add-device_function' as a processor for an output.

This is my first Golang project and I'm not a developer, but maybe we can get this up to shape and give gNMIc support for plugins together.

A couple of outstanding questions:
- Is there any better way to initialize the plugin_manager, now it uses init() and import in all.go.
- Add support for configuration of the plugin_manager, like debug log true/false and configurable plugin directory.
- Is there a way to send the logger object over RPC?

### DIFF
--- a/pkg/formatters/event_plugin/rpc.go
+++ b/pkg/formatters/event_plugin/rpc.go
@@ -76,7 +76,7 @@ func (g *EventProcessorRPC) Init(cfg interface{}, opts ...formatters.Option) err
 	}
 	err := g.client.Call("Plugin.Init", &InitArgs{Cfg: cfg}, &InitResponse{})
 	if err != nil {
-		panic(err)
+		return err
 	}
 	return nil
 }
@@ -85,7 +85,8 @@ func (g *EventProcessorRPC) Apply(event ...*formatters.EventMsg) []*formatters.E
 	var resp ApplyResponse
 	err := g.client.Call("Plugin.Apply", &ApplyArgs{Events: event}, &resp)
 	if err != nil {
-		panic(err)
+		log.Print("RPC client call error: ", err)
+		return nil
 	}
 	return resp.Events
 }
@@ -93,21 +94,21 @@ func (g *EventProcessorRPC) Apply(event ...*formatters.EventMsg) []*formatters.E
 func (g *EventProcessorRPC) WithActions(act map[string]map[string]interface{}) {
 	err := g.client.Call("Plugin.WithActions", act, &Actionresponse{})
 	if err != nil {
-		panic(err)
+		log.Print("RPC client call error: ", err)
 	}
 }
 
 func (g *EventProcessorRPC) WithTargets(tcs map[string]*types.TargetConfig) {
 	err := g.client.Call("Plugin.WithTargets", tcs, &Targetresponse{})
 	if err != nil {
-		panic(err)
+		log.Print("RPC client call error: ", err)
 	}
 }
 
 func (g *EventProcessorRPC) WithProcessors(procs map[string]map[string]any) {
 	err := g.client.Call("Plugin.WithProcessors", procs, &Proccessorresponse{})
 	if err != nil {
-		panic(err)
+		log.Print("RPC client call error: ", err)
 	}
 }
 
--- a/pkg/formatters/event_plugin/rpc.go
+++ b/pkg/formatters/event_plugin/rpc.go
@@ -1,0 +1,115 @@
+package event_plugin
+
+import (
+	"encoding/gob"
+	"log"
+	"net/rpc"
+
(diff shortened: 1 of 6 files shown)


karimra:
This is great, thanks for giving it a go.

> * Is there any better way to initialize the plugin_manager, now it uses init() and import in all.go.

That's the way

> * Add support for configuration of the plugin_manager, like debug log true/false and configurable plugin directory.

This will be required yes, I can look into it.

> * Is there a way to send the logger object over RPC?

I'm not sure that's possible, but what can be sent is the logger level (info, debug, etc...) and let the plugin decide where/how to log

pkg/formatters/event_plugin/rpc.go
```diff
@@ -0,0 +1,115 @@
...
+
+type EventProcessorRPC struct{ client *rpc.Client }
+
+func (g *EventProcessorRPC) Init(cfg interface{}, opts ...formatters.Option) error {
+	for _, opt := range opts {
+		opt(g)
+	}
+	err := g.client.Call("Plugin.Init", &InitArgs{Cfg: cfg}, &InitResponse{})
+	if err != nil {
+		panic(err)
```
karimra, line 79: maybe just return the error here instead of panic-ing

pkg/formatters/event_plugin/rpc.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return nil
+}
+
+func (g *EventProcessorRPC) Apply(event ...*formatters.EventMsg) []*formatters.EventMsg {
+	var resp ApplyResponse
+	err := g.client.Call("Plugin.Apply", &ApplyArgs{Events: event}, &resp)
+	if err != nil {
+		panic(err)
```
karimra, line 88: better log the error and return nil instead of panic

pkg/formatters/event_plugin/rpc.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return resp.Events
+}
+
+func (g *EventProcessorRPC) WithActions(act map[string]map[string]interface{}) {
+	err := g.client.Call("Plugin.WithActions", act, &Actionresponse{})
+	if err != nil {
+		panic(err)
```
karimra, line 96: better log the error and return instead of panic

pkg/formatters/event_plugin/rpc.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+}
+
+func (g *EventProcessorRPC) WithTargets(tcs map[string]*types.TargetConfig) {
+	err := g.client.Call("Plugin.WithTargets", tcs, &Targetresponse{})
+	if err != nil {
+		panic(err)
```
karimra, line 103: better log the error and return nil instead of panic

pkg/formatters/event_plugin/rpc.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+}
+
+func (g *EventProcessorRPC) WithProcessors(procs map[string]map[string]any) {
+	err := g.client.Call("Plugin.WithProcessors", procs, &Proccessorresponse{})
+	if err != nil {
+		panic(err)
```
karimra, line 110: better log the error and return nil instead of panic

karimra (changes requested):

karimra:
@fredsod can you allow edits on this PR for maintainers so I can fix the conflicts ? Thanks

fredsod:
> @fredsod can you allow edits on this PR for maintainers so I can fix the conflicts ? Thanks

Edits are allowed, you should be able to add commits to this PR

karimra:
It looks like you resolved the conflicts, thanks.

karimra (approved):

fredsod:
> It looks like you resolved the conflicts, thanks.

Have you had time to look into how to add support for configuration to the plugin_.manager?

karimra:
@fredsod: Yes, check #341 
I modified a few parts of the plugin manager to make it embeddable in the root app.
I also moved the plugin initialization to be part of the event-plugin init, that way the same plugin can be used under different processors of the same type. Each processor will get a separate running plugin.
The cleanup of the running plugins was not always happening, fixed that too.
Have a look at it and let me know what you think.

PR #317, opened 2023-12-14
Add topic-prefix for kafka output

mwdomino:
This PR adds a `topic-prefix` to the Kafka Output. When present, this prefix will be used to generate a separate Kafka topic per source. The topic name will follow `prefix_<source>`, and if there is a colon separating a port number as part of the source the colon will also be replaced with an underscore like `prefix_<source>_<port>` due to naming restrictions in Kafka topics.

I'm currently using this branch for testing in local development environments and all works as expected.

Let me know if you have any changes you'd like to ensure this keeps with the conventions used throughout the rest of the application, and thanks for the awesome work on gnmic as always!

### DIFF
--- a/docs/user_guide/outputs/kafka_output.md
+++ b/docs/user_guide/outputs/kafka_output.md
@@ -21,9 +21,9 @@ outputs:
     topic: telemetry 
     # Kafka topic prefix
     # If supplied, overrides the `topic` key and outputs to a separate topic per source
-    # named like `topic_<source>`. If `source` contains a port number separated with a colon,
+    # named like `$topic_$subscriptionName_$targetName`. If `source` contains a port number separated with a colon,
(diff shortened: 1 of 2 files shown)


karimra:
Thanks for adding this option to the Kafka output.
Do you mind adding the subscription name as well? Just like it's done on NATS output [here](https://github.com/openconfig/gnmic/blob/main/pkg/outputs/nats_outputs/nats/nats_output.go#L409)

mwdomino:
Great suggestion, I'll make the updates!

karimra:
I meant adding it to the Kafka output as well, you would end up with `$prefix_$subscriptionName_$targetName`

mwdomino:
I've updated this to capture the new formatting I believe. Let me know if you see anything else that you'd like modified.

karimra (approved):
Thanks for the contribution and for making the changes!

PR #320, opened 2023-12-15
Update docker-compose.yaml

i4networks:
influxdb 2.0 is not compatible with username password downgrade for sample to 1.8


karimra:
@i4networks if you don't mind, can you please sign the CLA?

Issue #319, opened 2023-12-15
what is the expected behavior poll mode and sync_response ?


amamory:
We are writing a gNMI server which implements the subscription POLL mode. 
The specification says when the server receives a poll subscription, it should send the Update message w data and the sync_response. Then, when the poll msg arrives, the process is repeated: send the Update message w data and the sync_response.
If the updates_only flag is true, after the subscription, only the sync_response is sent. the Update msg sent for the 1st time is skipped.

So, if you write a server that meets this interpretation, we see some buffering effect when using gnmic as a client. When we send the Poll message with gnmic, the data we receive is an old one. It's like the data has been buffered.  We added two screenshots to show it. The the 1st one, you see when we sent the poll msg, the latest value is 9 or 10, but the client receives 6. In the 2nd image, the latest value is 18 or 19, but it receives 10.

![IMG_20231215_114030](https://github.com/openconfig/gnmic/assets/8387109/5f5d178d-6cd1-4321-9f18-5d0c072dba8a)
![IMG_20231215_114048](https://github.com/openconfig/gnmic/assets/8387109/b38b9e8a-4885-4dd3-8a02-5af2c84808de)

Now, if you write a server that **never sends the sync_response message** (not when the subscription is received and also not when the Poll msg is receives), gnmic works beautifully ... meaning, gnmic receives the latest version of the data ... not a buffered data. **My conclusion is:** that gnmic client is not expecting the server to send sync_responses, although the specs says they should be sent.

So the questions are:
1) Perhaps our interpretation of the usage of sync_response in the poll mode is wrong ?!?!?
2) If the interpretation is correct, then why gnmic works only w a server that doesn't send sync_response messages ?

karimra:
Your interpretation is correct, gNMIc's POLL subscriptions handling was written when the requirement for sending sync_response after each set of update was not part of the spec yet.
This behavior needs to be fixed.

amamory:
thanks 👍👍👍
looking forward for this update !
it's a great tool!

karimra:
@amamory v0.35.0 has better support for POLL mode, please try it and let me know.

amamory:
I confirm the issue is solved. thank you very much ! great job !!!

PR #331, opened 2023-12-29
refactor poll subscription handling

karimra:
ph1 for #319

### DIFF
--- a/pkg/target/subscribe.go
+++ b/pkg/target/subscribe.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/jhump/protoreflect/dynamic"
 	"github.com/openconfig/gnmi/proto/gnmi"
+	"github.com/openconfig/gnmic/pkg/types"
 )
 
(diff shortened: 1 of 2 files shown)


hellt (approved):

pkg/target/subscribe.go
```diff
@@ -53,6 +54,7 @@ SUBSC:
 	t.subscribeCancelFn[subscriptionName] = cancel
 	subConfig := t.Subscriptions[subscriptionName]
```
vincentbernat, line 55: `subConfig` can be nil.

pkg/target/subscribe.go
```diff
@@ -264,3 +239,77 @@ func (t *Target) StopSubscription(name string) {
...
+	for {
+		if ctx.Err() != nil {
+			return nil
+		}
+		response, err := stream.Recv()
+		if err != nil {
+			return err
+		}
+		t.subscribeResponses <- &SubscribeResponse{
+			SubscriptionName:   subConfig.Name,
```
vincentbernat, line 270: Then, it crashes there.

PR #337, opened 2024-01-03
add subject.target.subscription format for Jetstream

mwdomino:
This PR adds an additional `subject` formatting option for Jetstream. It uses the existing `subject` key in the config and orders the subject like: `$subject.$target_name.$subscription_name`.

I'm using this in development already with no issues.

### DIFF
--- a/pkg/outputs/nats_outputs/jetstream/jetstream_output.go
+++ b/pkg/outputs/nats_outputs/jetstream/jetstream_output.go
@@ -61,6 +61,7 @@ type subjectFormat string
 
 const (
 	subjectFormat_Static                = "static"
+	subjectFormat_SubjectTargetSub      = "subject.target.subscription"
 	subjectFormat_SubTarget             = "subscription.target"
 	subjectFormat_SubTargetPath         = "subscription.target.path"
 	subjectFormat_SubTargetPathWithKeys = "subscription.target.pathKeys"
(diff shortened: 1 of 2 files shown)


pkg/outputs/nats_outputs/jetstream/jetstream_output.go
```diff
@@ -527,6 +529,17 @@ func (n *jetstreamOutput) subjectName(m proto.Message, meta outputs.Meta) (strin
...
+		sb.WriteString(n.Cfg.Subject)
+		sb.WriteString(".")
+		err := n.targetTpl.Execute(sb, meta)
+		if err != nil {
+			return "", err
+		}
+		sb.WriteString(".")
+		if sub, ok := meta["subscription-name"]; ok {
+			sb.WriteString(sub)
+		}
```
karimra, line 542: This assumes that a `subject` is always present.
Can we change this format to `target.subscription` and include `subject` only if `!= ""` ?
I can update the other formats to follow the same principle.

karimra (changes requested):

pkg/outputs/nats_outputs/jetstream/jetstream_output.go
mwdomino: That's a good idea, I'll update it.

karimra (approved):
Thanks for the contribution!

PR #345, opened 2024-01-10
Update sarama producer from sync to async

alibresco:
I've been seeing many messages dropped with error `timeout` when running the kafka output with high load, even with many kafka workers. I believe this is due to the interaction between the SyncProducer and the timeout mechanism in the code. Each call to `producer.SendMessage()` blocks the goroutine, and when all goroutines are waiting, messages time out in `k.msgChan`.

This PR switches from the sarama SyncProducer to AsyncProducer. This separates processing of kafka messages from the actual sending to kafka. This greatly improves the efficiency of sending messages to kafka because it allows the library to batch and compress messages together. This matches what the [sarama docs](https://github.com/IBM/sarama/blob/d9abf3cfb14d41fbc059d0f36195603000b1aad2/sync_producer.go#L9) say. Note that the configured timeout is still in effect because we pass it to the producer config.

In my testing, this PR completely eliminated kafka drops (which were previously about 40% of all messages) and also decreased the data rate sent to kafka by ~30% due to better compression.

### DIFF
--- a/pkg/outputs/kafka_output/kafka_output.go
+++ b/pkg/outputs/kafka_output/kafka_output.go
@@ -543,9 +543,7 @@ func (k *kafkaOutput) createConfig() (*sarama.Config, error) {
 	}
 
 	cfg.Producer.Retry.Max = k.cfg.MaxRetry
-	if k.cfg.SyncProducer {
-		cfg.Producer.Return.Successes = true
-	}
+	cfg.Producer.Return.Successes = true
 	cfg.Producer.Timeout = k.cfg.Timeout
 	switch k.cfg.RequiredAcks {
 	case requiredAcksNoResponse:
--- a/pkg/outputs/kafka_output/kafka_output.go
+++ b/pkg/outputs/kafka_output/kafka_output.go
@@ -280,13 +280,13 @@ func (k *kafkaOutput) RegisterMetrics(reg *prometheus.Registry) {
 }
 
 func (k *kafkaOutput) worker(ctx context.Context, idx int, config *sarama.Config) {
-	var producer sarama.SyncProducer
+	var producer sarama.AsyncProducer
 	var err error
 	defer k.wg.Done()
 	workerLogPrefix := fmt.Sprintf("worker-%d", idx)
 	k.logger.Printf("%s starting", workerLogPrefix)
 CRPROD:
-	producer, err = sarama.NewSyncProducer(strings.Split(k.Cfg.Address, ","), config)
+	producer, err = sarama.NewAsyncProducer(strings.Split(k.Cfg.Address, ","), config)
 	if err != nil {
 		k.logger.Printf("%s failed to create kafka producer: %v", workerLogPrefix, err)
 		time.Sleep(k.Cfg.RecoveryWaitTime)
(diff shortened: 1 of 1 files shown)


pkg/outputs/kafka_output/kafka_output.go
```diff
@@ -280,20 +280,61 @@ func (k *kafkaOutput) RegisterMetrics(reg *prometheus.Registry) {
...
+				if k.Cfg.Debug {
+					k.logger.Printf("%s failed to send a kafka msg to topic '%s': %v", workerLogPrefix, err.Msg.Topic, err.Err)
+				}
+				if k.Cfg.EnableMetrics {
+					kafkaNumberOfFailSendMsgs.WithLabelValues(config.ClientID, "send_error").Inc()
+				}
+			}
+		}
+	}()
+
```
karimra, line 337: These 2 goroutines can be just one. Have one `for` loop that selects on the 3 channels: `ctx.Done()`, `producer.Successes()` and `producer.Errors()`

alibresco: Makes sense. Fixed.

karimra:
I think it makes sense to still support sync producers via a config knob, there are advantages to both of them.
If you can do it, it would be great. Otherwise I can do it on top of this PR.

alibresco:
Feel free to go for it. I'm not sure I see the cleanest way to do it.

karimra:
@alibresco I added a knob to select the type of producer `sync-producer: true` will create a sync producer. The default will be an async producer.
I also added an parameter to choose the required acks, it used to default to wait for all, now it can be set to `no-response`, `wait-for-local` or `wait-for-all`. The default will be `wait-for-local`, it should help a bit with performance if there are replicas.
It would be great of you can test the changes on your side.

pkg/outputs/kafka_output/kafka_output.go
```diff
@@ -280,16 +296,125 @@ func (k *kafkaOutput) RegisterMetrics(reg *prometheus.Registry) {
 }
 
 func (k *kafkaOutput) worker(ctx context.Context, idx int, config *sarama.Config) {
+	if k.cfg.SyncProducer {
+		k.syncProducerWorker(ctx, idx, config)
+	}
+	k.asyncProducerWorker(ctx, idx, config)
```
alibresco, line 302: Should there be an `else` here?

pkg/outputs/kafka_output/kafka_output.go
```diff
@@ -398,32 +523,39 @@ func (k *kafkaOutput) createConfig() (*sarama.Config, error) {
...
 	}
 
-	cfg.Producer.Retry.Max = k.Cfg.MaxRetry
-	cfg.Producer.RequiredAcks = sarama.WaitForAll
+	cfg.Producer.Retry.Max = k.cfg.MaxRetry
 	cfg.Producer.Return.Successes = true
-	cfg.Producer.Timeout = k.Cfg.Timeout
+	cfg.Producer.Timeout = k.cfg.Timeout
+	switch k.cfg.RequiredAcks {
+	case requiredAcksNoResponse:
```
alibresco, line 549: Should this case set the sarama config to `NoResponse`? It looks like `WaitForLocal` is the sarama default.

alibresco:
@karimra Cool, I tested this against load with the default configuration options and confirmed no drops.

I left a couple questions about the code, but I don't think any of them affect the config options I'm using. I'm going to be on vacation next week, so feel free to merge this whenver you're happy with it.

Thanks for all your help!

pkg/outputs/kafka_output/kafka_output.go
karimra: Right, those worker functions are blocking calls. I should add a return after the syncProducerWorker() call. I try to avoid `else` statements.

pkg/outputs/kafka_output/kafka_output.go
karimra: Yes I wanted to follow the sarama default in this case.

karimra (approved):

PR #351, opened 2024-01-13
Don't use time.Sleep before retry while the context could be cancelled

vincentbernat:


karimra:
I don't think using time.Timer is a good idea here. The timer fires only once and needs to be reset in order to fire again. Calling reset on a timer can be tricky. 
Using a time.Ticker is not really the same as retrying.

Maybe use a time.After (not ideal) or create a new Timer for each retry and stop it if the ctx is canceled?

vincentbernat:
My initial attempt was mostly correct, but I moved the creation of timer to workaround a goto error. I pushed another tentative that I assume is now correct. The problem with time.After is that we leak a reference on each retry (since this is a long-running function).

karimra (approved):

karimra:
Thanks for fixing this!

PR #352, opened 2024-01-13
Don't rely on subscription configuration to get subscription name

vincentbernat:
When using the API directly, there is not subscription configuration (it is set in the app module) and we cannot use it to get the name. Instead, use the name provided in the function call.

This is a regression introduced in #331.

I am a bit unsure if this is the only regression as my tests are now failing. I am just quickly firing this one so you don't work on my comments in #331 in parallel.


vincentbernat:
This PR is now OK for me.

karimra (approved):

karimra:
Thanks for the contribution!

Issue #367, opened 2024-01-31
gnmic 0.35.1 - failed to create a gRPC client for target XXX: context deadline exceeded


tiago-amado:
# Description

1. gnmictls setup with 0.34.3 working fine for SROS
2. After upgrade to 0.35.1 its returning an error:

```bash
[root@AF6 CLAB_CA]# gnmic -a 172.20.20.4:57400 -u admin -p admin --tls-ca ca.pem capabilities
target "172.20.20.4:57400", capabilities request failed: failed to create a gRPC client for target "172.20.20.4:57400" : 172.20.20.4:57400: context deadline exceeded
Error: one or more requests failed
```

3. Downgrade to the previous release solves the error.

# Outputs

1. - Ok
```bash
[root@AF6 CLAB_CA]# gnmic version
version : 0.34.3
commit : d461fa9
   date : 2023-11-14T02:13:11Z
gitURL : https://github.com/openconfig/gnmic
   docs : https://gnmic.openconfig.net/
[root@AF6 CLAB_CA]# 
[root@AF6 CLAB_CA]# 
[root@AF6 CLAB_CA]# gnmic -a 172.20.20.4 -u admin -p admin --tls-ca ca.pem capabilities
gNMI version: 0.8.0
supported models:
  - nokia-conf, Nokia, 23.10.R2
  - nokia-state, Nokia, 23.10.R2
  - nokia-li-state, Nokia, 23.10.R2
supported encodings:
  - JSON
  - BYTES
  - PROTO
  - JSON_IETF

[root@AF6 CLAB_CA]#
```

2. Upgrade NOK
```bash
[root@AF6 ~]#
[root@AF6 ~]# bash -c "$(curl -sL https://get-gnmic.openconfig.net/)"
gnmic 0.35.1 is available. Changing from version 0.34.3.
Downloading https://github.com/openconfig/gnmic/releases/download/v0.35.1/gnmic_0.35.1_linux_x86_64.tar.gz
Preparing to install gnmic 0.35.1 into /usr/local/bin
gnmic installed into /usr/local/bin/gnmic
version : 0.35.1
commit : 91f4205
   date : 2024-01-25T21:10:47Z
gitURL : https://github.com/openconfig/gnmic
   docs : https://gnmic.openconfig.net/
[root@AF6 ~]# 
[root@AF6 ~]# 
[root@AF6 ~]# gnmic version
version : 0.35.1
commit : 91f4205
   date : 2024-01-25T21:10:47Z
[...]

PR #371, opened 2024-02-05
explicitly set default TLS cipher suite

karimra:
#367

### DIFF
--- a/pkg/utils/tls.go
+++ b/pkg/utils/tls.go
@@ -27,6 +27,40 @@ import (
 	"time"
 )
 
+// from https://github.com/golang/go/blob/b39ec942d8f154add0af01ebf66d7524318470e2/src/crypto/tls/cipher_suites.go#L270
+// we explicitly set the default ciphers so that go-grpc doesn't filter out forbidden ciphers: https://github.com/grpc/grpc-go/pull/6776
+// https://github.com/openconfig/gnmic/issues/367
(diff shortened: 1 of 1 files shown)


hellt (approved):

vincentbernat:
Are you sure this is is equivalent to before? This is enabling ciphers that are deemed insecure since a long time. There is tls.CipherSuites() which has more ciphers but is still avoiding the insecure ones. 3DES and RC4 should never be used anymore. Which cipher SROS was using that was put on the insecure list? Maybe an AES CBC one?

karimra:
This is just to solve #367, working on making the cipher suites configurable. 
SROS only supports RSA as key exchange algorithm.

vincentbernat:
`tls.CipherSuites()` also includes RSA key exchange. That was the default value before I think. The one you took are only used for ordering. Nobody keeps RC4 in the default list of ciphers.

karimra:
I understand, like I said this was a fix for a specific issue to unblock someone. There will be another PR where the default cipher suites chosen will better mimic what the stdlib does.
`tls.CipherSuites()` does not return what the stdlib selects by default (it includes both TLS1.2 and 1.3 and does not follow the preferred order), it's a bit more complicated than that, see [here](https://github.com/golang/go/blob/go1.21.6/src/crypto/tls/cipher_suites.go#L339)

PR #406, opened 2024-04-01
Sort prometheus labels in accordance with the prometheus spec

alibresco:
Currently, the gnmic prometheus output does not sort labels lexicographically before sending them out the door.

This is required by the prometheus [remote write spec](https://prometheus.io/docs/concepts/remote_write_spec/):

> MUST have label names sorted in lexicographical order.

It's unclear how all implementations of prometheus remote write targets handle the unsorted labels. In at least some, it can lead to issues with correctness or sharding efficiency.

I have made the fix and added a test.

### DIFF
--- a/pkg/outputs/prometheus_output/prometheus_common.go
+++ b/pkg/outputs/prometheus_output/prometheus_common.go
@@ -9,10 +9,12 @@
 package prometheus_output
 
 import (
+	"cmp"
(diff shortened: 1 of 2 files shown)


karimra (approved):

karimra:
Thanks for the contribution!

PR #412, opened 2024-04-18
Add a config option for kafka producer flush frequency

alibresco:
Sarama exposes [a config option](https://github.com/IBM/sarama/blob/4ad35041300e1c15ba58b630745ac8eb05f30c10/config.go#L253) for the best effort frequency of flushes.

This option tells the library to hold kafka messages for up to this amount of time with the goal of better client-side compression and batching. This can lead to peformance improvements in both gnmic and on the kafka broker.

By default, sarama sets this value to 0 which indicates that it should flush as frequently as possible. This PR preserves that behavior while adding a knob to increase the frequency.


karimra (approved):

karimra:
Thanks for the contribution

PR #450, opened 2024-05-22
add gNMI commit extension

karimra:


jarrodb:
Nice, Karim!

pkg/app/set.go
```diff
@@ -139,6 +139,12 @@ func (a *App) InitSetFlags(cmd *cobra.Command) {
 	cmd.Flags().StringArrayVarP(&a.Config.LocalFlags.SetUpdateCli, "update-cli", "", []string{}, "a cli command to be sent as a set update request")
 	cmd.Flags().StringVarP(&a.Config.LocalFlags.SetUpdateCliFile, "update-cli-file", "", "", "path to a file containing a list of commands that will be sent as a set update request")
 
+	cmd.Flags().StringVarP(&a.Config.LocalFlags.SetCommitId, "commit-id", "", "", "commit ID value")
+	cmd.Flags().BoolVarP(&a.Config.LocalFlags.SetCommitConfirm, "confirm", "", false, "confirm the commit ID")
+	cmd.Flags().BoolVarP(&a.Config.LocalFlags.SetCommitRequest, "request", "", false, "start a commit confirmed transaction")
+	cmd.Flags().BoolVarP(&a.Config.LocalFlags.SetCommitCancel, "cancel", "", false, "cancel the commit")
```
hellt, line 145: I think I would have an easier read on the flags meaning if they would be

- commit-confirm
- commit-request
- commit-cancel

otherwise the CLI looks ambigious to me:

`gnmic -a test set --update-path /abc --update-value /val1 --request`

the `request` alone just didn't map in my head with commit request.

jarrodb: I agree with Roman is because it conforms to the protobuf definition and would be more explicit

```
message Commit {
  string id = 1;
  oneof action {
    CommitRequest commit = 2;
    CommitConfirm confirm = 3;
    CommitCancel cancel = 4;
    CommitSetRollbackDuration set_rollback_duration = 5;
```

karimra: Will update, readability > conciseness.

hellt (approved):

PR #461, opened 2024-06-13
api support to update target subscriptions directly

jarrodb:
This PR introduces a new API endpoint `/config/targets/:id/subscriptions` designed to allow users to directly update the subscription list for a specific target without the need to delete and re-add the target.

The use case for this endpoint allows us to build an initial target list with a single subscription for system information (i.e. `eos_native:/Eos/image` for EOS or `/system/information/version` for SRL) that we can update with subscriptions for specific paths that vary across vendor versions to accommodate breaking changes in models.

### DIFF
--- a/docs/user_guide/api/targets.md
+++ b/docs/user_guide/api/targets.md
@@ -220,7 +220,6 @@ Returns an empty body if successful.
     ```json
     {
         "errors": [
-            "subscriptions not found",
             "subscription $subscription does not exist"
         ]
     }
--- a/docs/user_guide/api/targets.md
+++ b/docs/user_guide/api/targets.md
@@ -194,3 +194,42 @@ Returns an empty body if successful.
         ]
     }
     ```
+    
+## `PATCH /api/v1/targets/{id}/subscriptions`
+
+Updates existing subscriptions for the target ID
+    
+Returns an empty body if successful.
+
+=== "Request"
+    ```bash
+    curl --request PATCH gnmic-api-address:port/api/v1/targets/192.168.1.131:57400/subscriptions -d '{"subscriptions": ["sub1", "sub2"]}'
+    ```
+=== "200 OK"
+    ```json
+    ```
+=== "404 Not found"
+    ```json
+    {
+        "errors": [
+            "target $target not found"
+        ]
+    }
+    ```
+=== "400 Bad Request"
+    ```json
+    {
+        "errors": [
+            "subscriptions not found",
+            "subscription $subscription does not exist"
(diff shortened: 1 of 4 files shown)


pkg/app/target.go
```diff
@@ -96,6 +95,26 @@ func (a *App) DeleteTarget(ctx context.Context, name string) error {
 	return nil
 }
 
+// UpdateTargetConfig updates the subscriptions for an existing target
+func (a *App) UpdateTargetSubscription(ctx context.Context, name string, subs []string) error {
+	a.configLock.Lock()
+	for _, subName := range subs {
+		if _, ok := a.Config.Subscriptions[subName]; !ok {
+			return fmt.Errorf("subscription %q does not exist", subName)
```
karimra, line 103: The config lock will not be unlocked if this error is returned.

jarrodb: woops.. started with a defer in a previous iteration and ended up changing the logic.  nice catch!

docs/user_guide/api/targets.md
```diff
@@ -193,4 +193,43 @@ Returns an empty body if successful.
...
+        ]
+    }
+    ```
+=== "400 Bad Request"
+    ```json
+    {
+        "errors": [
+            "subscriptions not found",
+            "subscription $subscription does not exist"
+        ]
```
karimra, line 225: Looking at `UpdateTargetSubscription`, I don't see 2 errors being returned at the same time. I think these are 2 distinct responses.

jarrodb: right.. i wasn't sure if this was trying to capture all potential 400 responses or just serve as an example of a single bad response, so i opted for all 🤷

jarrodb: i can adjust the example to return a single response

karimra: I think it makes sense to adjust it if you don't mind.

jarrodb: of course dude!

jarrodb: i _think_ that solves it.. if you want i can work on adding some real tests to the api side

karimra: nice, tests would be great.

karimra (approved):

karimra:
Thanks for contribution!

jarrodb:
Oh I was going to get a test pushed first (I'll still do that)

karimra:
Sorry, I understood you will do it in a separate PR

PR #481, opened 2024-07-11
Added a no-trim flag for SET operations

rao-aneesh:
We can send CLI commands using the `--update-path cli: --update-file <path-to-file>` args (replace and union-replace as well) by keeping CLI commands in the input file. The files are always trimmed when read. This causes an issue if the encoding is ASCII and the router expects the CLI command to terminate with a '\n'.

Using the `--update-cli-file` arg does solve this issue (since the file isn't trimmed in this case) but it locks the origin into 'cli'. From a testing perspective, I want to try 'xr_cli' or 'junos_cli'. I would also like to test negative cases where origin is not cli. Also, `union-replace-cli` is not a flag that exists, hence this change.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -156,6 +156,7 @@ type LocalFlags struct {
 	SetRequestVars            string        `mapstructure:"set-request-vars,omitempty" json:"set-request-vars,omitempty" yaml:"set-request-vars,omitempty"`
 	SetRequestProtoFile       []string      `mapstructure:"set-proto-request-file,omitempty" yaml:"set-proto-request-file,omitempty" json:"set-proto-request-file,omitempty"`
 	SetDryRun                 bool          `mapstructure:"set-dry-run,omitempty" json:"set-dry-run,omitempty" yaml:"set-dry-run,omitempty"`
+	SetNoTrim                 bool          `mapstructure:"set-no-trim,omitempty" json:"set-no-trim,omitempty" yaml:"set-no-trim,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -685,9 +686,13 @@ func (c *Config) CreateSetRequest(targetName string) ([]*gnmi.SetRequest, error)
 				c.logger.Printf("error reading data from file '%s': %v", c.LocalFlags.SetUpdateFile[i], err)
 				return nil, err
 			}
+			trimChars := " \r\n\t"
+			if c.LocalFlags.SetNoTrim {
+				trimChars = ""
+			}
```
karimra, line 692: it would make sense to declare `trimChars` as a const and reuse it in this file.
it also makes more sense to set the variable to the empty string first them set it to `trimChars` if `!NoTrim`

karimra (changes requested):

pkg/config/config.go
rao-aneesh: If `trimChars` is declared as a const, we won't be able to change its value right? Do you want to use a 2nd variable?

karimra: Right, declare a const that gets assigned to a variable if needed.

rao-aneesh: Changed it. If this is what you requested, can you tell me why this is a better implementation?

karimra: The value is reused at multiple occasions, it's better to declare it once and reuse it.

karimra (approved):

PR #482, opened 2024-07-12
Added print-request and dry-run support for subscribe

rao-aneesh:
Added print-request and dry-run support for subscribe

### DIFF
--- a/pkg/app/gnmi_client_subscribe.go
+++ b/pkg/app/gnmi_client_subscribe.go
@@ -211,9 +211,21 @@ CRCLIENT:
 	a.Logger.Printf("target %q gNMI client created", t.Config.Name)
 
 	for _, sreq := range subRequests {
+		if a.Config.GlobalFlags.PrintRequest || a.Config.LocalFlags.SubscribeDryRun {
+			err := a.PrintMsg(tc.Name, "Subscribe Request:", sreq.req)
+			if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/app/gnmi_client_subscribe.go
```diff
@@ -211,10 +211,22 @@ CRCLIENT:
...
+		if a.Config.LocalFlags.SubscribeDryRun {
+			continue
+		}
 		a.Logger.Printf("sending gNMI SubscribeRequest: subscribe='%+v', mode='%+v', encoding='%+v', to %s",
 			sreq.req, sreq.req.GetSubscribe().GetMode(), sreq.req.GetSubscribe().GetEncoding(), t.Config.Name)
 		go t.Subscribe(gnmiCtx, sreq.req, sreq.name)
 	}
+	if a.Config.LocalFlags.SubscribeDryRun {
+		os.Exit(0)
+	}
```
karimra, line 229: In the case the subscribe command is running against multiple targets, the os.Exit(0) will stop the process before all the subscribe requests are printed.

karimra (changes requested):

pkg/app/gnmi_client_subscribe.go
rao-aneesh: After the return statement, the program is left waiting for something. I was not able to figure out where it waits. Can you help me out here?

PR #496, opened 2024-07-30
Add shutdown route

mwdomino:
Howdy!

This PR adds a simple route `/api/v1/admin/shutdown` which cancels the `App` main context and shuts down the running `gnmic` server. 

In our environment we run `gnmic` in k8s and have an in-memory datastore. When we detect that this datastore has failed/crashed for some reason we would like a way to signal `gnmic` to restart immediately to trigger resubscription/resync of all paths.

Let me know if you have any feedback on this or have another approach you'd recommend.

### DIFF
--- a/docs/user_guide/api/other.md
+++ b/docs/user_guide/api/other.md
@@ -16,3 +16,14 @@ Health check endpoint for Kubernetes or similar
         "status": "healthy"
     }
     ```
+    
+## /api/v1/admin/shutdown
+
+### `POST /api/v1/admin/shutdown`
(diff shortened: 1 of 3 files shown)


karimra:
Thanks for the contribution. 
The expectation here is that the gNMIc container stops and it gets restarted by k8s ?

mwdomino:
Yes, exactly. Just shutting down is good enough for us as we can rely on k8s to spawn a new pod.

karimra (approved):

PR #497, opened 2024-07-31
add support for rotating file outputs

mwdomino:
This PR adds support for rotating file outputs, similar to log rotation. In our internal use of gnmic I have written some scripts to replay production events logged from the `file` output into a development environment so that we can reproduce bugs and do some load testing. Whenever I needed a production dump I would restart gnmic with a file output enabled for an hour or so and then grab the file and roll back to our previous configuration.

This option enables us to limit the disk space and just keep the last 3x 100MB files available on our k8s nodes should we need to grab them to replay or troubleshoot. No behavior is changed if not using the `rotating` file-type.

You can specify the file-type as `rotating` and then supply the following configuration to configure the rotation parameters:
```yaml
    rotation:
      max-size: 100 # size in megabytes
      max-age: 30 # max age in days
      max-backups: 3 # maximum number of old files to store, not counting the current file
      compress: false # whether or not to enable compression
```

This uses the `lumberjack` package in Go to handle the rotation when enabled. 

To support this I've converted the `File.file` from an `*os.File` to a new `file` interface that is satisfied by the new `rotatingFile` struct. This allowed me to add the new behavior without modifying any of the existing logic.

Let me know if you have any comments or concerns about this!

### DIFF
--- a/pkg/outputs/file/file_output.go
+++ b/pkg/outputs/file/file_output.go
@@ -157,7 +157,7 @@ func (f *File) Init(ctx context.Context, name string, cfg map[string]interface{}
 			return err
 		}
 
-		f.file = newRotatingFile(f.cfg.FileName, f.cfg.Rotation.Compress, f.cfg.Rotation.MaxSize, f.cfg.Rotation.MaxBackups, f.cfg.Rotation.MaxAge)
+		f.file = newRotatingFile(f.cfg)
 	default:
 	CRFILE:
 		f.file, err = os.OpenFile(f.cfg.FileName, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
--- a/pkg/outputs/file/file_output.go
+++ b/pkg/outputs/file/file_output.go
@@ -78,7 +78,7 @@ type Config struct {
 	EnableMetrics      bool           `mapstructure:"enable-metrics,omitempty"`
 	Debug              bool           `mapstructure:"debug,omitempty"`
 	CalculateLatency   bool           `mapstructure:"calculate-latency,omitempty"`
-	Rotation           RotationConfig `mapstructure:"rotation,omitempty"`
+	Rotation           rotationConfig `mapstructure:"rotation,omitempty"`
 }
 
 type file interface {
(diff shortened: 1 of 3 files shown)


pkg/outputs/file/rotating_file.go
```diff
@@ -0,0 +1,64 @@
...
+	"gopkg.in/natefinch/lumberjack.v2"
+)
+
+// RotationConfig manages configuration around file rotation
+type RotationConfig struct {
+	MaxSize    int  `mapstructure:"max-size,omitempty"`
+	MaxBackups int  `mapstructure:"max-backups,omitempty"`
+	MaxAge     int  `mapstructure:"max-age,omitempty"`
+	Compress   bool `mapstructure:"compress,omitempty"`
+}
```
karimra, line 15: I think this can be unexported, `rotatingConfig`

karimra (changes requested):

pkg/outputs/file/rotating_file.go
```diff
@@ -0,0 +1,64 @@
...
+
+	if rc.MaxBackups == 0 {
+		return fmt.Errorf("Rotation.MaxBackups is required if using type 'rotating'")
+	}
+
+	if rc.MaxAge == 0 {
+		return fmt.Errorf("Rotation.MaxAge is required if using type 'rotating'")
+	}
+
+	return nil
```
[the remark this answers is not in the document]
mwdomino: This is a great call and I have no idea why I didn't just do that initially...

mwdomino:
Thanks for the feedback, I've made those changes and pushed the updates :)

pkg/outputs/file/file_output.go
```diff
@@ -142,6 +149,8 @@ func (f *File) Init(ctx context.Context, name string, cfg map[string]interface{}
 		f.file = os.Stdout
 	case "stderr":
 		f.file = os.Stderr
+	case "rotating":
+		f.file = newRotatingFile(f.cfg)
```
karimra, line 153: I wonder if it makes more sense to define `Rotation` as a pointer and handle it in the `default:` case.
If `cfg.Rotation != nil`, then create a rotating file (after setting the missing fields to their defaults).

mwdomino: I think that makes a lot more sense, and makes the config a bit easier to understand. Updated!

karimra (approved):

karimra:
Thanks for the contribution!

Issue #503, opened 2024-08-08
Support for the gNMI extension field


Ichabond:
gNMI extensions (https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-extensions.md) allow for extended features in the gRPC messages sent back.
The extension field is present at the top-level of the gNMI RPCs.
Juniper as a vendor is registering an extensionID for their telemetry: https://github.com/openconfig/gnmi/pull/76
Right now, their prototext looks like this:
```
extension: {
  registered_ext: {
    id: 1
    msg: "\n\x0cedge01.lis01\x10\xff\xff\x03\"\x0fsensor_1007_2_1*d/interfaces/interface[name='irb']/subinterfaces/subinterface[index='100']/state/counters/out-octets/2d/interfaces/interface[name='irb']/subinterfaces/subinterface[index='100']/state/counters/out-octets/:\x05mib2d@\x01H\xff\x89\x99\x97\x932P\xff\x89\x99\x97\x932`\x81\x8a\x99\x97\x932p\x01\x80\x01\x02"
  }
}
```
which contains a lot of helpful data, which should be exported as tags on the events.

The ask:
Add generic support for the extension field, allowing for arbitrary protobuf files to be used to decode extension IDs (a config map of ID -> protobuf file maybe?). Then use the field in processor to generate tags from it.

The specific reason for this ask:
This extension field contains linecard information of the linecard generating the response. This is necessary data to deduplicate certain events (logical interfaces spanning multiple linecards will generate an event per linecard, which will be clobbered, given they look like duplicates right now). If the extension data was exported as tags, it would be a unique event.

Ichabond:
Specifically, this is an extension of the existing `--proto-file` support, but which is locked to the Nokia SROS protofiles.

I'll have a look if this could be done by just letting the user define a mapping of:
extensionID -> protofile + proto root.

This would then get decoded properly, I think, given I'm seeing `Unmarshall` errors on the JSON side which are related to the binary-marshalled protobuf in the `msg` field inside `extension`.
So I _think_ if the code here: https://github.com/openconfig/gnmic/blob/main/pkg/api/target/subscribe.go#L290 is made more generic, at least decoding would work.
From there, I think it being converted to tags would "just work"?
Will try to get some PoC going to prove/disprove this understanding.

Ichabond:
@karimra I have a working prototype.
Before I make a PR though, a couple questions I'd appreciate your input on:
[...]

PR #509, opened 2024-08-14
Add gNMI Extension field parsing support

Ichabond:
gNMI allows for the use of an `extension` field in each top-level message of the gNMI RPCs: https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-extensions.md Given this is an arbitrary Protobyte payload, the default gNMI protobufs can't decode the payload contained within the field. This PR adds the necessary configuration options to load in an arbitrary Protobuf file per `extension` identifier, with a message-name to lookup the message type.

In this change:
1. A new function `DecodeExtension` is added. This function uses `protoreflect` to dynamically marshal arbitrary protoBytes into JSON. The loaded JSON is then put back in the Extension message as bytes (mainting type)
2. The `Target` type has an `ExtensionProtoMap` added, allowing for the lookup of Extension IDs to loaded-in protobufs
3. The required changes to the `TargetConfig` type to support loading in the new configuration
4. Modified `collector.go` to output the gNMI message _after_ inlining the decoded protoBytes
5. Loading in the protobufs was added to `app/target.go`: `parseExtensionProtos`. This uses the `Parser` provided by `protoreflect/desc/protoparse`
6. Added functionality to `event.go` to insert the K/Vs provided by the Extension as Tags. Given we come from JSON, all numbers are float64, so the only 2 types supported currently are `string` and `float64`
7. Minor helper function to turn the arbitrary JSON into an arbitrary map.

This has been tested with a device emiting an `extension` field:
```
[gnmic] target "edge01_test01": gNMI Subscribe Response: &{
SubscriptionName:port_stats
SubscriptionConfig:{"name":"port_stats","paths":["/interfaces/interface/"],"mode":"STREAM","stream-mode":"SAMPLE","encoding":"JSON","sample-interval":15000000000,"heartbeat-interval":15000000000,"outputs":["prom-scrape-output"]}
Response:
update:{timestamp:1723653363502452302  prefix:{elem:{name:"interfaces"}  elem:{name:"interface"  key:{key:"name"  value:"et-1/0/3"}}}
update:{path:{elem:{name:"state"}  elem:{name:"hardware-port"}}  val:{json_val:"\"FPC1:PIC0:PORT3\""}}
update:{path:{elem:{name:"state"}  elem:{name:"transceiver"}}  val:{json_val:"\"FPC1:PIC0:PORT3:Xcvr0\""}}}
extension:{registered_ext:{id:1
    msg:"{\"systemId\":\"edge01_test01\",\"componentId\":65535,\"sensorName\":\"sensor_1005_2_1\",\"subscribedPath\":\"/interfaces/interface/\",\"streamedPath\":\"/interfaces/interface/\",\"component\":\"chassisd\",\"sequenceNumber\":\"770002\",\"payloadGetTimestamp\":\"1723653363502\",\"streamCreationTimestamp\":\"1723653361858\",\"exportTimestamp\":\"1723653363504\",\"streamId\":\"PERIODIC\"}"}}}
```
Which is then properly rendered to a Prometheus metric:
```
gnmi_interfaces_interface_state_hardware_port{component="chassisd",componentId="65535",hardware_port="FPC1:PIC0:PORT3",interface_name="et-1/0/3",metric_source="edge01_test01",subscription_name="port_stats",systemId="edge01_test01"} 1
```
Note that some label-drop rules have been added to remove the spurious labels to avoid a cardinality explosion.

### DIFF
--- a/pkg/api/target/subscribe.go
+++ b/pkg/api/target/subscribe.go
@@ -303,6 +303,25 @@ func (t *Target) DecodeProtoBytes(resp *gnmi.SubscribeResponse) error {
 	return nil
 }
 
+func (t *Target) DecodeExtension(resp *gnmi.SubscribeResponse) error {
+	if t.ExtensionProtoMap == nil {
+		return nil
+	}
+	for _, extension := range resp.Extension {
+		m := dynamic.NewMessage(t.ExtensionProtoMap[int(extension.GetRegisteredExt().GetId().Number())])
+		err := m.Unmarshal(extension.GetRegisteredExt().GetMsg())
+		if err != nil {
+			return err
+		}
+		jsondata, err := m.MarshalJSON()
+		if err != nil {
+			return err
+		}
+		extension.GetRegisteredExt().Msg = jsondata
+	}
+	return nil
+}
+
 func (t *Target) DeleteSubscription(name string) {
 	t.m.Lock()
 	defer t.m.Unlock()
--- a/pkg/app/collector.go
+++ b/pkg/app/collector.go
@@ -67,13 +67,21 @@ func (a *App) StartCollector(ctx context.Context) {
 				select {
 				case rsp := <-rspChan:
 					subscribeResponseReceivedCounter.WithLabelValues(t.Config.Name, rsp.SubscriptionConfig.Name).Add(1)
-					if a.Config.Debug {
-						a.Logger.Printf("target %q: gNMI Subscribe Response: %+v", t.Config.Name, rsp)
+					// decode gNMI extensions
+					if extensions := rsp.Response.Extension; len(extensions) > 0 {
+						err := t.DecodeExtension(rsp.Response)
+						if err != nil {
+							a.Logger.Printf("target %q: failed to decode extension field: %v", t.Config.Name, err)
+							continue
+						}
 					}
 					err := t.DecodeProtoBytes(rsp.Response)
 					if err != nil {
 						a.Logger.Printf("target %q: failed to decode proto bytes: %v", t.Config.Name, err)
 						continue
+					}
+					if a.Config.Debug {
+						a.Logger.Printf("target %q: gNMI Subscribe Response: %+v", t.Config.Name, rsp)
 					}
 					m := outputs.Meta{
 						"source":            t.Config.Name,
--- a/pkg/app/target.go
+++ b/pkg/app/target.go
@@ -14,6 +14,9 @@ import (
 
 	"github.com/fullstorydev/grpcurl"
 
+	"github.com/jhump/protoreflect/desc"
+	"github.com/jhump/protoreflect/desc/protoparse"
+
 	"github.com/openconfig/gnmic/pkg/api/target"
 	"github.com/openconfig/gnmic/pkg/api/types"
 )
@@ -36,6 +39,10 @@ func (a *App) initTarget(tc *types.TargetConfig) (*target.Target, error) {
 			}
 		}
 		err := a.parseProtoFiles(t)
+		if err != nil {
+			return nil, err
+		}
+		err = a.parseExtensionProtos(t)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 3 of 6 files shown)


Ichabond:
This is a proposed solution for #503

Ichabond:
Updated the commit to fix the type assertion issue as well as the variable assignment issue.

Ichabond:
Anything else needed here @karimra ?

pkg/app/target.go
```diff
@@ -155,6 +162,34 @@ func (a *App) parseProtoFiles(t *target.Target) error {
...
 }
 
+// Dynamically parse (and load) protobuf files defined in config for specific extension IDs
+func (a *App) parseExtensionProtos(t *target.Target) error {
+	parser := protoparse.Parser{}
+	extensionProtoMap := make(map[int]*desc.MessageDescriptor)
+	a.Logger.Printf("Target %q loading protofiles for gNMI extensions", t.Config.Name)
+	if len(t.Config.RegisteredExtensions) == 0 {
+		return nil
+	}
```
karimra, line 172: I would move this to the beginning of the function

pkg/app/collector.go
```diff
@@ -67,14 +67,22 @@ func (a *App) StartCollector(ctx context.Context) {
...
+						}
 					}
 					err := t.DecodeProtoBytes(rsp.Response)
 					if err != nil {
 						a.Logger.Printf("target %q: failed to decode proto bytes: %v", t.Config.Name, err)
 						continue
 					}
+					if a.Config.Debug {
+						a.Logger.Printf("target %q: gNMI Subscribe Response: %+v", t.Config.Name, rsp)
+					}
```
karimra, line 85: Can you move this back under the case statement to that the debug logging always happens regardless of decoding succeeding or not ?

pkg/formatters/event.go
```diff
@@ -40,9 +42,28 @@ func ResponseToEventMsgs(name string, rsp *gnmi.SubscribeResponse, meta map[stri
...
+			}
+			for k, v := range extensionValues {
+				switch v := v.(type) {
+				case string:
+					prefixTags[k] = v
+				case float64:
+					prefixTags[k] = strconv.FormatFloat(v, 'G', -1, 64)
+				}
+			}
+		}
```
karimra, line 66: This part assumes that any extension data can and will be mapped to a map[string]any, and that the goal is to get the extension data as tags, it will not always be the case for all extensions.

pkg/api/target/subscribe.go
```diff
@@ -303,6 +303,25 @@ func (t *Target) DecodeProtoBytes(resp *gnmi.SubscribeResponse) error {
...
+		m := dynamic.NewMessage(t.ExtensionProtoMap[int(extension.GetRegisteredExt().GetId().Number())])
+		err := m.Unmarshal(extension.GetRegisteredExt().GetMsg())
+		if err != nil {
+			return err
+		}
+		jsondata, err := m.MarshalJSON()
+		if err != nil {
+			return err
+		}
+		extension.GetRegisteredExt().Msg = jsondata
```
karimra, line 320: I'm not sure it's a good idea to overwrite the extension Msg with the jsondata here. It makes more sense to return the resulting json and pass it on to the outputs as metadata.

karimra (changes requested):

karimra:
While this works for that junos extension, I'm not sure it will for any other extension.
The data in an extension is not always meant to be added as tags, we should find a way to pass it on to the output and use processors to select some keys to set as tags.
The other thing I'm not entirely convinced by is that these proto files are set by target. If you have 100s of targets of the same type (same extensions), we will have to parse the proto files 100s of times. This is fine for now, but it should be improved/fixed in a future PR.

pkg/app/collector.go
```diff
@@ -67,14 +67,22 @@ func (a *App) StartCollector(ctx context.Context) {
...
+						if err != nil {
+							a.Logger.Printf("target %q: failed to decode extension field: %v", t.Config.Name, err)
+							continue
+						}
 					}
 					err := t.DecodeProtoBytes(rsp.Response)
 					if err != nil {
 						a.Logger.Printf("target %q: failed to decode proto bytes: %v", t.Config.Name, err)
 						continue
 					}
```
karimra, line 82: We shouldn't drop the subscribe response if we fail to decode the extension.

peejaychilds:
Hello awesome people.   Any movement on this.   It would great to be able to do some processing of the attributes from extensions in for example starlark - so we could filter/markup/combine etc as needed.

triggerhappy17:
Very interested in this also. Any update ?

coredump17:
@Ichabond  are you still progressing this ?  It would be a great feature

PR #532, opened 2024-10-22
Add a metric for failed subscriptions

netixx:
This adds a way to monitor failed subscription response received, it makes it easier to identified missconfigured devices or subscriptions.

Maybe we should include the error in the label as well ?

### DIFF
--- a/pkg/app/collector.go
+++ b/pkg/app/collector.go
@@ -118,7 +118,8 @@ func (a *App) StartCollector(ctx context.Context) {
 				case tErr := <-errChan:
 					if errors.Is(tErr.Err, io.EOF) {
 						a.Logger.Printf("target %q: subscription %s closed stream(EOF)", t.Config.Name, tErr.SubscriptionName)
-					} else {
+						} else {
+						subscribeResponseFailedCounter.WithLabelValues(t.Config.Name, tErr.SubscriptionName).Inc()
 						a.Logger.Printf("target %q: subscription %s rcv error: %v", t.Config.Name, tErr.SubscriptionName, tErr.Err)
 					}
 					if remainingOnceSubscriptions > 0 {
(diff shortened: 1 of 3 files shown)


pkg/app/metrics.go
```diff
@@ -28,6 +28,13 @@ var subscribeResponseReceivedCounter = prometheus.NewCounterVec(prometheus.Count
 	Help:      "Total number of received subscribe response messages",
 }, []string{"source", "subscription"})
 
+var subscribeResponseFailedCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+	Namespace: "gnmic",
+	Subsystem: "subscribe",
+	Name:      "number_of_failed_subscribe_response_messages_total",
```
karimra, line 34: Let's call this `number_of_failed_subscribe_request_messages_total`, the error is received at the gRPC level not in a SubscribeResponse message.

pkg/app/metrics.go
```diff
@@ -28,6 +28,13 @@ var subscribeResponseReceivedCounter = prometheus.NewCounterVec(prometheus.Count
...
+	Help:      "Total number of failed subscribe requsts",
```
karimra, line 35: `requsts` -> `requests`

pkg/app/collector.go
```diff
@@ -116,6 +116,7 @@ func (a *App) StartCollector(ctx context.Context) {
 						return
 					}
 				case tErr := <-errChan:
+					subscribeResponseFailedCounter.WithLabelValues(t.Config.Name, tErr.SubscriptionName).Inc()
```
karimra, line 119: Let's move the `.Inc()` into the `else` statement. `EOF` is not a real error, it's how a server closes a stream.

karimra (changes requested):

pkg/app/metrics.go
netixx: done

pkg/app/collector.go
netixx: done

pkg/app/metrics.go
netixx: fixed

karimra:
Thanks for the contribution and for making the changes!

PR #550, opened 2024-11-06
fix NATS output handling of empty messages

karimra:


pkg/outputs/nats_outputs/jetstream/jetstream_output.go
```diff
@@ -490,8 +490,8 @@ func (n *jetstreamOutput) createNATSConn(c *config) (*nats.Conn, error) {
 		nats.ErrorHandler(func(_ *nats.Conn, _ *nats.Subscription, err error) {
 			n.logger.Printf("NATS error: %v", err)
 		}),
-		nats.DisconnectHandler(func(*nats.Conn) {
-			n.logger.Println("Disconnected from NATS")
+		nats.DisconnectErrHandler(func(c *nats.Conn, err error) {
```
hellt, line 493: do we need `c` receiver in this anon func? seems to not be used in the func body

hellt (approved):

pkg/outputs/nats_outputs/jetstream/jetstream_output.go
karimra: removed

PR #556, opened 2024-11-16
Do not return targets passwords in REST API requests

karimra:


nlgotz (approved):
Looks good to me! Thanks for doing this update so quickly!

hellt (approved):

Issue #563, opened 2024-12-11
Loaders don't detect changes in target data


netixx:
I stumbled upon an issue where one of my targets changed IP address, and gnmic was still using the old IP for telemetry.
I am using HTTP loader every 5 minutes, so if a devices changes IP between 2 fetch, I would like gnmic to update the target and use the new IP.
It also applies to other target fields (such as tags).

Currently only adds and deletes are taken into account, not the changes inside the target.

As per this function:
```
func Diff(m1, m2 map[string]*types.TargetConfig) *TargetOperation {
	result := &TargetOperation{
		Add: make(map[string]*types.TargetConfig, 0),
		Del: make([]string, 0),
	}
	if len(m1) == 0 {
		for n, t := range m2 {
			result.Add[n] = t
		}
		return result
	}
	if len(m2) == 0 {
		for name := range m1 {
			result.Del = append(result.Del, name)
		}
		return result
	}
	for n, t := range m2 {
		if _, ok := m1[n]; !ok {
			result.Add[n] = t
		}
	}
	for n := range m1 {
		if _, ok := m2[n]; !ok {
			result.Del = append(result.Del, n)
		}
	}
	return result
}
```

I am not sure what is the best approach,
I was thinking that the easiest way to fix that would be to artificially generate a change (remove the old target, and add the updated target), this way all the hooks trigger and we can start fresh ?

karimra:
That is by design, it has to be an explicit delete then add.
artificially generate a delete if a target change would work. 
The tricky part is figuring out if a change requires entirely recreating the target.

netixx:
OK I see,
but this means that if a target is updated in any way (connection parameters or variables),
I need to first remove it from the list, and then add it.

I am using the HTTP loader, so it seems really cumbersome to do that (I guess I would need to add an intermediate software that detect changes in the targets, removes changed targets and serve them back on the next query from gnmic, but even if I do that, the changed target will not be polled until the loader runs again).

How can we streamline the integration of target changes for dynamic loader ?

PR #603, opened 2025-01-29
Handle target changes in loaders

netixx:
When a target config changes,
generate a delete and a add action
so that the subscription can be restarted
to apply the new parameters.

https://github.com/openconfig/gnmic/issues/563


karimra (approved):

netixx:
Thanks!

karimra:
Thanks for the contribution!

PR #608, opened 2025-02-20
TargetUpMetric shouldn't show deleted targets

### LINKED ISSUE #604: When a target is deleted, the metric doesn't get removed
When using Consul or HTTP loaders, removing a device results in the `gnmic_target_up{name="router"}` metric always being `1` even though the device is no longer subscribed.

Looking at the issue, it has been over 2 hours since the router was removed from Consul and the metric is still showing as a 1.

## Steps to reproduce:

1. Setup GNMIC to use Consul loader for targets
2. Start subscribing to devices
3. Check the `/metrics` path and see that the `gnmic_target_up` metrics look good
4. Remove a device from Consul
5. Check the `/metrics` path and see that the deleted target still shows as a `1`.

nlgotz:
Fixes: #604 

This resets the TargetUpMetric before creating the metrics so that deleted targets (tested via Consul) don't show as still up.

### DIFF
--- a/pkg/app/metrics.go
+++ b/pkg/app/metrics.go
@@ -89,7 +89,7 @@ func (a *App) registerTargetMetrics() {
 						ownTargets[strings.TrimPrefix(k, lockedNodesPrefix+"/")] = v
 					}
 				}
-
+				targetUPMetric.Reset()
 				a.configLock.RLock()
 				for _, tc := range a.Config.Targets {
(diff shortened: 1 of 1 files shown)


yawningORsnarling (approved):
GTG

karimra (approved):

karimra:
Thanks for the contribution!

PR #610, opened 2025-02-24
feat: Templated names and tags for Consul targets

nlgotz:
This adds the ability to use Go Templates on Consul targets to set both the name of the target as well as event-tags. The primary driving force for this comes from Telegraf's [Prometheus Input Plugin](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/prometheus) which also has Consul support and allows for using Go templates to set the name/tags as well.

As an example, you are now able to do something like:

```yaml
loader:
    type: consul
    address: consul:8500
    datacenter: "dc99"
    services:
        - name: gnmic-sros
          target-name: "{{.Meta.device}}"
          target-tags:
              region: "{{.Meta.region}}"
              geo: "{{.Meta.geo}}"
              test: "{{.Meta.test}}"
```

This also removes a debugging statement in `pkg/config/loader.go` that could potentially expose passwords in log messages

### DIFF
--- a/docs/user_guide/targets/target_discovery/consul_discovery.md
+++ b/docs/user_guide/targets/target_discovery/consul_discovery.md
@@ -25,9 +25,9 @@ loader:
 
 ### Templating with Consul
 
-It is possible to set the target name to something other than the Consul Service ID using the `target-name` field under the service. The target name can be customized using [Go Templates](https://golang.org/pkg/text/template/).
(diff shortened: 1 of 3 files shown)


karimra:
Why not use the existing fields under `config`? They can be set to some Go template, the only missing piece is to execute them.

karimra (approved):

karimra:
Thanks for the contribution! And sorry for the delay.

Issue #633, opened 2025-04-05
`gnmic path --with-prefix` should use the module name as a prefix


hellt:
The `--with-prefix` argument of the `gnmic path` command outputs paths with yang prefixes, literally.

Yet, what it needs to use as a prefix is the module name, and not the module prefix, as the latter has local significance.

For example, a path generated for SR Linux with `--with-prefix`:

```
 	
/srl_nokia-if:interface[name=*]/traffic-rate
```

shows the usage of the `srl_nokia-if` as a prefix, while it should've used the [module name](https://github.com/nokia/srlinux-yang-models/blob/v25.3/srlinux-yang-models/srl_nokia/models/interfaces/srl_nokia-interfaces.yang#L1) instead, like:

```
 	
/srl_nokia-interfaces:interface[name=*]/traffic-rate
```

I think the keeping the `--with-prefix` flag name is OK, and we can change the value only.

PR #635, opened 2025-04-05
replace module prefix with module name when generating xpaths with prefix

karimra:
#633


hellt (approved):
thanks, looks good
CleanShot 2025-04-05 at 21 03 17@2x

PR #674, opened 2025-06-10
implement subscribe-backoff for loaders and config change events

elobommart:
When new targets are subscribed to via a loader (HTTP, file ..) or via a config change event, the subscribe backoff is not implemented.

Therefore, it is not possible to precisely smooth out the load when loading a lot of targets via loaders and using sample subscriptions.


karimra:
Thanks for the contribution!

PR #685, opened 2025-07-10
gNMI server: add timeout config option

ShimmerGlass:
gNMIc's gNMI server has a hardcoded 2min timeout on unary requests. While this value is fine for most cases it causes issues in our shop for long-ish cli origin requests.
This change makes this timeout configurable via `gnmi-server.timeout` in the config. The default value remains 2min, behavior is kept unchanged unless `gnmi-server.timeout` is explicitly set.


ShimmerGlass:
Hello, any news ? :)

PR #690, opened 2025-07-20
Fix sync_response output for ONCE mode subscriptions

earies:
https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#35151-once-subscriptions

Current behavior was suppressing `sync_response` output for ONCE mode
subscriptions due to lack of switch/case fallthrough


earies:
@karimra bump for review/comment

PR #725, opened 2025-09-11
feat: add pprof support

senthilsam:
A much-needed one for debugging when something went wrong. I was able to use this to trace memory leaks 
 
 
image

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -111,6 +111,9 @@ type App struct {
 	tunTargetCfn  map[tunnel.Target]context.CancelFunc
 	// processors plugin manager
 	pm *plugin_manager.PluginManager
+
+	// pprof
+	pprof *PprofServer
 }
 
 func New() *App {
@@ -149,6 +152,9 @@ func New() *App {
 		ttm:          new(sync.RWMutex),
 		tunTargets:   make(map[tunnel.Target]struct{}),
 		tunTargetCfn: make(map[tunnel.Target]context.CancelFunc),
+
+		// pprof
+		pprof: NewPprofServer(),
 	}
 	a.router.StrictSlash(true)
 	a.router.Use(headersMiddleware, a.loggingMiddleware)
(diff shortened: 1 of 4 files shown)


pkg/app/pprof.go
```diff
@@ -0,0 +1,52 @@
...
+		if len(parts) == 2 && parts[0] == "" {
+			pprofHostPort = "localhost:" + parts[1]
+		}
+		pprofHostPort = "http://" + pprofHostPort + "/debug/pprof"
+
+		fmt.Printf("I! Starting pprof HTTP server at: %s\n", pprofHostPort)
+
+		server := &http.Server{
+			Addr:         address,
+			ReadTimeout:  10 * time.Second,
```
karimra, line 39: 10 seconds for WriteTimeout is short, leave it unset or set it to 1 minute or higher.
Pprof does not expect a request body, set `ReadHeaderTimeout` to 5 or 10 seconds and leave ReadTimeout unset.

pkg/app/const.go
```diff
@@ -14,6 +14,7 @@ const (
 	defaultGrpcPort   = "57400"
 	msgSize           = 512 * 1024 * 1024
 	defaultRetryTimer = 10 * time.Second
+	defaultPprofAddr  = "localhost:6060"
```
karimra, line 17: use 127.0.0.1 instead of localhost

pkg/app/pprof.go
```diff
@@ -0,0 +1,52 @@
...
+		err: make(chan error),
+	}
+}
+
+func (p *PprofServer) Start(address string) {
+	go func() {
+		pprofHostPort := address
+		parts := strings.Split(pprofHostPort, ":")
+		if len(parts) == 2 && parts[0] == "" {
+			pprofHostPort = "localhost:" + parts[1]
```
karimra, line 31: use 127.0.0.1 instead of localhost

karimra (changes requested):

pkg/app/app.go
```diff
@@ -218,6 +227,20 @@ func (a *App) InitGlobalFlags() {
 }
 
 func (a *App) PreRunE(cmd *cobra.Command, args []string) error {
+	if a.Config.EnablePprof {
+		_, _, err := net.SplitHostPort(a.Config.GlobalFlags.PprofAddr)
+		if err != nil {
+			fmt.Printf("pprof error %v", err)
```
karimra, line 233: return the error here, you don't need to print it out.

karimra (changes requested):

PR #727, opened 2025-09-11
configurable grpc buffer

senthilsam:


karimra:
@senthilsam could you address the open comments ?

PR #737, opened 2025-09-15
feat: enhance target up metric

senthilsam:
Currently target_up metric is 1 even when the connection failed because of auth issues

So I have made changes to reflect with grpc connection states, with one minor change, even from grpc Ready=2, i have kept ready as 1 so our users will still see target_up =1 when the target is actually up
```
const (
	// Idle indicates the ClientConn is idle.
	Idle [State](https://pkg.go.dev/google.golang.org/grpc/connectivity#State) = [iota](https://pkg.go.dev/builtin#iota)
	// Connecting indicates the ClientConn is connecting.
	Connecting
	// Ready indicates the ClientConn is ready for work.
	Ready
	// TransientFailure indicates the ClientConn has seen a failure but expects to recover.
	TransientFailure
	// Shutdown indicates the ClientConn has started shutting down.
	Shutdown
)
```

### DIFF
--- a/pkg/app/metrics.go
+++ b/pkg/app/metrics.go
@@ -114,13 +114,16 @@ func (a *App) registerTargetMetrics() {
 							targetUPMetric.WithLabelValues(tc.Name).Set(1)
 							targetConnStateMetric.WithLabelValues(tc.Name).Set(1)
 						case "CONNECTING":
+							targetUPMetric.WithLabelValues(tc.Name).Set(0)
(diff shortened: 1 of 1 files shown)


karimra:
I'm not sure it's a good idea to modify the existing metric. The intent behind it is to indicate if a target "exists" or not.
Instead I would add a new metric called `target_connection_state` that reflects the different gRPC connection states.

senthilsam:
thanks, made the changes

senthilsam:
@karimra can you please review this.
Also, when is the next release planned??

PR #739, opened 2025-09-17
Add option to enable profiling (pprof)

netixx:
pprof is exposed on standard routes on the API server along with the metrics.
It enables the possibility to do troublshooting of CPU/RAM usage (eg. memory-leak analysis).

Note that I had to bump the go version in the dockerfile,
because the go.mod has a 1.24 requirement.

1.25 is probably ambitious, also you'll maybe want to handle that in another PR, so happy to edit this one.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ on:
   #     - "!releases/**"
   
 env:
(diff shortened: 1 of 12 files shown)


karimra (approved):

karimra:
Thanks for the contribution!

PR #771, opened 2025-11-03
fix #770 The API path to patch subscriptions for a target ID is wrong

### LINKED ISSUE #770: The API path to patch subscriptions for a target ID is documented incorrectly
As per this [code](https://github.com/openconfig/gnmic/blob/6b35566f4f266898fb8be67bc4958e287de11a42/pkg/app/routes.go#L43) the path to patch subscriptions for a target ID is `/config/targets/{id}/subscriptions`. But, this is documented incorrectly as `/targets/{id}/subscriptions` [here](https://github.com/openconfig/gnmic/blob/6b35566f4f266898fb8be67bc4958e287de11a42/docs/user_guide/api/targets.md?plain=1#L198).

I had tested the same in my setup and I'm getting a 404 when I tried to PATCH using the path `/targets/{id}/subscriptions`. The same worked when I tried using the path `/config/targets/{id}/subscriptions`.

ashvanth48t:


karimra:
Thanks for the contribution!

PR #780, opened 2025-12-04
Add OTLP output for OpenTelemetry Protocol integration

drewelliott:
## Summary

Adds native OpenTelemetry Protocol (OTLP) output support to gNMIc, enabling direct export of telemetry data to OTLP-compatible backends.

- New OTLP output plugin with direct OTLP/gRPC export
- Full metric conversion (gauges, counters, histograms) from gNMI to OTLP
- Support for custom resource attributes and semantic conventions
- Comprehensive test coverage included

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/outputs/otlp_output/otlp_output.go
+++ b/pkg/outputs/otlp_output/otlp_output.go
@@ -385,9 +385,11 @@ func (o *otlpOutput) worker(ctx context.Context, id int) {
 	for {
 		select {
 		case <-ctx.Done():
-			// Flush remaining batch
+			// Flush remaining batch with fresh context to avoid sending with cancelled context
 			if len(batch) > 0 {
-				o.sendBatch(ctx, batch)
+				flushCtx, cancel := context.WithTimeout(context.Background(), o.cfg.Timeout)
+				defer cancel()
+				o.sendBatch(flushCtx, batch)
 			}
 			if o.cfg.Debug {
 				o.logger.Printf("worker %d stopped", id)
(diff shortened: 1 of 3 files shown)


pkg/outputs/otlp_output/otlp_output.go
```diff
@@ -0,0 +1,488 @@
...
+			if len(batch) > 0 {
+				o.sendBatch(ctx, batch)
+				batch = make([]*formatters.EventMsg, 0, o.cfg.BatchSize)
+			}
+		}
+	}
+}
+
+// sendBatch sends a batch of events to the OTLP collector
+func (o *otlpOutput) sendBatch(ctx context.Context, events []*formatters.EventMsg) {
```
karimra: `ctx` is not not used any where in `sendBatch`, maybe you meant to pass it to `sendGRPC`? You could simplify it and just use `o.ctx` since it's stored in the receiver struct.

pkg/outputs/otlp_output/otlp_output.go
```diff
@@ -0,0 +1,488 @@
...
+		opts = append(opts, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
+	} else {
+		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	}
+
+	// Connect with timeout
+	ctx, cancel := context.WithTimeout(context.Background(), o.cfg.Timeout)
+	defer cancel()
+
+	conn, err := grpc.DialContext(ctx, o.cfg.Endpoint, opts...)
```
karimra: `grpc.DialContext` is deprecated, it's advised to use `grpc.NewClient`.
While it works now, it will fail and never retry (`Init` returns) if the OTLP endpoint is not reachable at the time the output tries to connect.
`grpc.NewClient` creates a client without attempting to dial. It will dial when the client sends the first RPC.

karimra (changes requested):

karimra (approved):

PR #785, opened 2025-12-19
add windows build

mcarbonneaux:
this pull request add windows build

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,8 +13,8 @@ on:
     tags:
       - v*
 env:
-  GOVER: 1.24.7
-  GORELEASER_VER: v1.19.2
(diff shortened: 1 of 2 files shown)


karimra:
Windows builds were remove some time ago: https://github.com/karimra/gnmic/pull/196
Any reason using WSL is not an option?
There are a few dependencies that don't work that well on Windows (go-prompt for example), that's why windows builds were removed.

mcarbonneaux:
test run ok : https://github.com/mcarbonneaux/gnmic/actions/runs/20915617555 
but macos and windows staticcheck fail, and i think is reviewdog/action-staticcheck but on this plateform (they download static binary of another plateform...), i've removed because they dont need to be executed on any architecture but on code, the linux one is suffisant.

PR #800, opened 2026-02-14
Added buffer and window size to dial options

Philldomd:
This contribution will add grcp dial options for each target.

They are configured per target in config file.


karimra:
Thanks for the contribution!

PR #862, opened 2026-05-04
feat: OTLP output map boolean values to int 1/0

moshe010:
Extend createNumberDataPointWithValue to handle bool values by mapping true → 1 and false → 0 as NumberDataPoint_AsInt. Previously booleans were treated as unsupported and dropped.

### DIFF
--- a/pkg/outputs/otlp_output/otlp_converter.go
+++ b/pkg/outputs/otlp_output/otlp_converter.go
@@ -317,6 +317,12 @@ func (o *otlpOutput) createNumberDataPointWithValue(cfg *config, event *formatte
 		} else {
 			return nil
 		}
+	case bool:
+		if v {
+			dp.Value = &metricspb.NumberDataPoint_AsInt{AsInt: 1}
+		} else {
+			dp.Value = &metricspb.NumberDataPoint_AsInt{AsInt: 0}
+		}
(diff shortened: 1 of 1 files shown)


karimra:
@moshe010 you need to sign the CLA for the PR to be reviewed and accepted.

moshe010:
> @moshe010 you need to sign the CLA for the PR to be reviewed and accepted.

thanks I think I fixed it now

pkg/outputs/otlp_output/otlp_converter.go
```diff
@@ -317,6 +317,12 @@ func (o *otlpOutput) createNumberDataPointWithValue(cfg *config, event *formatte
 		} else {
 			return nil
 		}
+	case bool:
+		if v == true {
```
karimra: This is nitpicking but it's more idiomatic to write `if v {` here instead of `if v == true {`. Otherwise, the rest LGTM

karimra (changes requested):

pkg/outputs/otlp_output/otlp_converter.go
moshe010: fixed

karimra (approved):

PR #863, opened 2026-05-04
Add per-metric string-value labels for prometheus output

tarko:
When a gNMI subscription emits string-typed leaves alongside numeric ones (e.g. clei-code, name, descriptive identifiers), the existing strings-as-labels option turns those strings into Prometheus labels and applies them to all metrics produced from that event. This pollutes unrelated metrics with labels they don't logically belong to.

In my case I'm dealing with existing metrics infrastructure that has some assumptions about where (and more specifically, where-not) labels are added. As far as I've investigated, there is currently no way to do it as none of the processing functions allows one to split out specific values (so event-value-tag etc could be used to attach tags only to those values).

This POC (not to be merged as-is) adds a new opt-in Prometheus output config option strings-as-single-metric-labels that converts string-typed values into a label attached only to the metric derived from that value, rather than cross-applying it to every metric in the same event.

Example

```
 {
   "name": "sub1",
   "tags": { "interface_name": "onu 1/1/95", "source": "lab-olt" },
   "values": {
     "/interfaces-state/interface/admin-status":  "1",
     "/interfaces-state/interface/name":          "onu 1/1/95",
     "/interfaces-state/interface/onu/clei-code": "CLEI-1234"
   }
 }
```

With strings-as-single-metric-labels: true:

```
 interfaces_state_interface_admin_status{interface_name="onu 1/1/95",source="lab-olt"} 1
 interfaces_state_interface_name{interface_name="onu 1/1/95",source="lab-olt",name="onu 1/1/95"} 1
 interfaces_state_interface_onu_clei_code{interface_name="onu 1/1/95",source="lab-olt",clei_code="CLEI-1234"} 1
```

If this is acceptable then I can rework this accordingly, add support in TimeSeriesFromEvent etc.

### DIFF
--- a/pkg/app/processor.go
+++ b/pkg/app/processor.go
@@ -135,11 +135,12 @@ func (a *App) promFormat(rrevs [][]*formatters.EventMsg, outName string) ([]byte
 		return nil, fmt.Errorf("output %q must be of type 'prometheus' or 'remote_write'", outName)
 	}
 	mb := &promcom.MetricBuilder{
-		Prefix:                 a.Config.FileConfig.GetString(outputPath + "/metric-prefix"),
-		AppendSubscriptionName: a.Config.FileConfig.GetBool(outputPath + "/append-subscription-name"),
-		StringsAsLabels:        a.Config.FileConfig.GetBool(outputPath + "/strings-as-labels"),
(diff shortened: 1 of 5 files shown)


karimra:
@tarko you need to sign the CLA for the PR to be reviewed and accepted.

tarko:
CLA completed

PR #864, opened 2026-05-07
fix: address Docker and Prometheus CVEs (GHSA-x744-4wpc-v9h2, GHSA-pxq6-2prw-chj9, GHSA-vffh-x6r8-xx99)

ronan-nexthop:
## Why

gnmic v0.45.0 carries three known CVEs in its dependency tree that downstream packagers cannot work around with `go get @latest`:

| CVE | Severity | Dependency | Fixed in |
|---|---|---|---|
| GHSA-x744-4wpc-v9h2 | HIGH | `github.com/docker/docker` | needs v29.3.1 (only available as `github.com/moby/moby/v2`) |
| GHSA-pxq6-2prw-chj9 | MEDIUM | `github.com/docker/docker` | same as above |
| GHSA-vffh-x6r8-xx99 | MEDIUM | `github.com/prometheus/prometheus` | needs v0.311.2+ |

The Docker fixes can't land via the existing `github.com/docker/docker` import path because Docker's v29 source code only ships under the renamed `github.com/moby/moby/v2` module — there is no v29.x.y tag at the old path. The Prometheus bump fails `go mod tidy` because it pulls in `k8s.io/kube-openapi`, whose transitive `github.com/go-openapi/swag/loading` test imports a since-renamed package (`testify/v2/assert/yaml`) that no longer exists.

## What

Three commits:

### 1. `fix(docker_loader): migrate from docker/docker to moby/moby/v2`

The migration is API-breaking, not just an import rewrite — `pkg/loaders/docker_loader/docker_loader.go` is the only consumer:

- The `filters` package was removed entirely. `filters.NewArgs / filters.Args / filters.KeyValuePair` are gone, replaced by `client.Filters` (a `map[string]map[string]bool` with chainable `Add`).
- `container.ListOptions` and `network.ListOptions` moved into the `client` package as `ContainerListOptions` and `NetworkListOptions`.
- `ContainerList` and `NetworkList` now return result structs rather than slices; callers must use `.Items` to get the underlying list.
- `Client.Ping` requires a `PingOptions` argument.
- `EndpointSettings.IPAddress` / `.GlobalIPv6Address` and `PortSummary.IP` are now `netip.Addr` instead of `string`. String comparisons against `""` / `"0.0.0.0"` / `"::"` become `IsValid()` / `IsUnspecified()` and a `String()` conversion at format-time.

After this change `github.com/docker/docker` is no longer in the module graph (`go mod why` reports "main module does not need package"; `go list -m all | grep docker/docker` is empty).

### 2. `chore(deps): bump prometheus to v0.311.2 to address GHSA-vffh-x6r8-xx99`

Bumps `github.com/prometheus/prometheus` past the CVE-fixed bar and pins `github.com/go-openapi/testify/enable/yaml/v2` and `github.com/go-openapi/testify/v2` to v2.5.0 (`// indirect`) so MVS picks the version where the upstream maintainers already corrected the broken import (renamed `testify/v2/assert/yaml` to `testify/v2/enable/stubs/yaml`).

After this commit `go mod tidy` succeeds.

### 3. `chore(ci): bump Go toolchain to 1.25.9`

The prometheus bump (and its transitive otel/k8s/x/* updates) raises the `go.mod` floor to `go 1.25.0`. CI workflows and the published Dockerfile still pinned `1.24.12`, which means every CI run wasted time auto-downloading 1.25 to satisfy the directive. This commit bumps:

- `.github/workflows/test.yml` `GOVER: 1.24.12 → 1.25.9`
- `.github/workflows/lint.yml` `GOVER: 1.24.12 → 1.25.9`
- `.github/workflows/release.yml` `GOVER: 1.24.12 → 1.25.9`
- `Dockerfile` `FROM golang:1.24.12 → golang:1.25.9`

## Verification

- `go mod tidy` — clean
- `go build ./...` — clean
- `go vet ./...` — clean
- **Smoke tested against a real Docker 29.4.1 daemon**: invoked the migrated `docker_loader.RunOnce` with two labelled `nginx:alpine` containers and confirmed it discovered both with correct addresses derived from published ports + `gnmic-port` labels (`172.17.0.2:57400`, `172.17.0.3:57401`).
- The unix-socket and bridge-fallback IP-resolution branches are exercised by structurally identical code paths but were not run in the smoke test — review of the `netip.Addr` rewrites would benefit from a second set of eyes.

## Risk / blast radius

- Anyone consuming `gnmic` as a library at Go ≥ 1.25 (or with the default `GOTOOLCHAIN=auto`, which auto-downloads 1.25) is unaffected; the public API of `pkg/loaders/docker_loader` is unchanged.
- Consumers running `GOTOOLCHAIN=local` on a Go 1.24 install will need to upgrade to Go 1.25 — but they would have hit the same wall the moment any of the bumped transitive deps (prometheus, otel, k8s.io/*, x/crypto, x/net, x/oauth2, etc.) reached their machine via MVS, regardless of gnmic's `go` directive. The bump is honest documentation of the floor that already exists.
- Anyone using the `docker` loader at runtime gets identical behavior — same target name, same address resolution rules.


pkg/loaders/docker_loader/docker_loader.go
```diff
@@ -363,11 +359,11 @@ func (d *dockerLoader) getTargets(ctx context.Context) (map[string]*types.Target
...
 							for _, nr := range nrs {
 								if n, ok := cont.NetworkSettings.Networks[nr.Name]; ok {
-									if n.IPAddress != "" {
-										tc.Address = n.IPAddress
+									if n.IPAddress.IsValid() {
+										tc.Address = n.IPAddress.String()
 										break
 									}
-									tc.Address = n.GlobalIPv6Address
+									tc.Address = n.GlobalIPv6Address.String()
```
karimra: `n.GlobalIPv6Address.String()` needs to be wrapped in a `if n.GlobalIPv6Address.IsValid() {}` as well I think

pkg/loaders/docker_loader/docker_loader.go
```diff
@@ -417,11 +416,11 @@ func (d *dockerLoader) getTargets(ctx context.Context) (map[string]*types.Target
...
 								for _, nr := range nrs {
 									if n, ok := cont.NetworkSettings.Networks[nr.Name]; ok {
-										if n.IPAddress != "" {
-											tc.Address = n.IPAddress
+										if n.IPAddress.IsValid() {
+											tc.Address = n.IPAddress.String()
 											break
 										}
-										tc.Address = n.GlobalIPv6Address
+										tc.Address = n.GlobalIPv6Address.String()
```
karimra: same as previous comment

karimra (changes requested):

pkg/loaders/docker_loader/docker_loader.go
ronan-nexthop: yeah good catch, without this the stringer will return a error string which wont catch the check on 374.

pkg/loaders/docker_loader/docker_loader.go
ronan-nexthop: resolved

ronan-nexthop:
merge conflicts in go.mod, resolved them, re-generated the go.mod/go.sum and re-ran the unit tests to confirm no regressions

karimra (approved):

ronan-nexthop:
thanks @karimra, can we get a new release publish that contains these fixes please

PR #868, opened 2026-05-12
Fix containerlab link

Bzhash:


hellt:
ty @Bzhash 
I have covered all occurrences of the old url and did a repo-wide sweep in #979

PR #885, opened 2026-07-04
fix: getTunnelTargetMatch never reports a match

### LINKED ISSUE #884: tunnel-target-matches never matches any tunnel target ("target ignored, not matching any rule")
### Description
Every gRPC-tunnel target registration is rejected with `"target ignored, not matching any rule"`,
regardless of any `tunnel-server.targets` entry or dynamically-applied `tunnel-target-matches`
config, even when the target's ID and Type exactly match a configured rule.

### Root cause
In `pkg/collector/managers/targets/tunnel_server.go`, `getTunnelTargetMatch()` passes a filter
closure to `store.List("tunnel-target-matches", ...)`. The closure correctly checks the ID regex,
the Type regex, and builds the resulting `TargetConfig` -- but every code path, including the
successful one, falls through to the unconditional `return false` at the end of the function
(https://github.com/openconfig/gnmic/blob/main/pkg/collector/managers/targets/tunnel_server.go#L241-L274).
Since `store.List`'s contract only includes a key in its result when the filter returns `true`
(see `github.com/zestor-dev/zestor/store/gomap.(*memStore).List`), `matchingConfigs` is always
empty, and `getTunnelTargetMatch` always returns `nil`.
This has been broken since the `tunnel-target-matches` feature was introduced in 7e297f7 ("add
tunnel-target-matches").

### Repro
1. Configure a `tunnel-server` with any `targets` match rule (or push one dynamically via the
   `tunnel-target-matches` store/API), e.g. `id: "^my-device$"`, `type: "GNMI_GNOI"`.
2. Connect a tunnel client that registers as `{ID: "my-device", Type: "GNMI_GNOI"}`.
3. Observe the log: `"target ignored, not matching any rule"`, even though the ID/Type match
   exactly.

### Fix
One-line fix: add `return true` at the end of the successful match branch inside the `switch`.
PR: #885

zentavr:
Fixes #884 

`getTunnelTargetMatch`'s filter closure (passed to `store.List`) checks the target ID regex,
Type regex, and builds the resulting `TargetConfig` - but every branch, including the
successful match, fell through to the unconditional `return false` at the end of the function.
`store.List` only includes an entry when the filter returns `true`, so `matchingConfigs` was
always empty and every tunnel target was rejected as `"target ignored, not matching any rule"`,
regardless of configuration.

This has been broken since the `tunnel-target-matches` feature was added in 7e297f7.
- One-line fix: `return true` on the successful match path.
- Added a regression test (`tunnel_server_test.go`) that fails against the old code and passes
  with the fix.

### DIFF
--- a/pkg/collector/managers/targets/tunnel_server.go
+++ b/pkg/collector/managers/targets/tunnel_server.go
@@ -269,6 +269,7 @@ func (ts *tunnelServer) getTunnelTargetMatch(tt tunnel.Target) *types.TargetConf
 				ts.logger.Error("failed to set target config defaults", "error", err, "id", tt.ID, "type", tt.Type)
 				return false
 			}
+			return true
 		}
 		return false
 	})
--- a/pkg/collector/managers/targets/tunnel_server_test.go
+++ b/pkg/collector/managers/targets/tunnel_server_test.go
@@ -1,0 +1,54 @@
+package targets_manager
+
+import (
+	"log/slog"
+	"testing"
+
+	"github.com/openconfig/gnmic/pkg/api/types"
+	"github.com/openconfig/gnmic/pkg/config"
+	"github.com/zestor-dev/zestor/store"
+	"github.com/zestor-dev/zestor/store/gomap"
+	"github.com/openconfig/grpctunnel/tunnel"
+)
+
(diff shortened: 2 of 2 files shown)


karimra (approved):

karimra:
Thanks for the contribution!

PR #887, opened 2026-07-05
fix(otlp): validate counter-patterns in subscribe mode

### LINKED ISSUE #880: OTLP output: `counter-patterns` ignored in subscribe mode - all metrics exported as Gauge
## Version Details
**gNMIc version:** v0.46.0
**Output:** `otlp_output` (gRPC)

## Summary

When gNMIc is run as a standard `subscribe` daemon, the OTLP output's `counter-patterns` option has no effect: every numeric metric is exported as a `Gauge`, even when the value key clearly matches a configured pattern. As a result, counters such as `/interfaces/interface/state/counters/out-octets` arrive downstream (e.g. Vector/Datadog) as gauges instead of monotonic cumulative Sums.

## Potential Root cause

`counter-patterns` are only consumed via the compiled `config.counterRegexes` slice, which is populated exclusively inside `validateConfig()`:

```go
func (o *otlpOutput) validateConfig(c *config) error {
    ...
    c.counterRegexes = make([]*regexp.Regexp, 0, len(c.CounterPatterns))
    for _, p := range c.CounterPatterns {
        re, err := regexp.Compile(p)
        ...
        c.counterRegexes = append(c.counterRegexes, re)
    }
    return nil
}
```

`validateConfig()` is called only from `Update()` and `Validate()`:

- `Init()` (the path used by `gnmic subscribe`) calls `setDefaultsFor()` and then `o.cfg.Store(ncfg)`, but **never calls `validateConfig()`**, and `setDefaultsFor()` does not compile the regexes.
- `Validate()` compiles the regexes on a **local throwaway `ncfg`** and returns without storing it, so it does not affect the config that `Init()` stores for runtime use.

Consequently, the running config has `counterRegexes == nil`, and `isCounter()` always returns `false`:

```go
func (o *otlpOutput) isCounter(cfg *config, valueName string) bool {
    for _, re := range cfg.counterRegexes { // nil in subscribe mode
        if re.MatchString(valueName) { return true }
    }
    return false
}
```

so `convertEventToMetrics()` takes the `Gauge` branch for every numeric value. The option only works when running in a mode that invokes `Update()` (dynamic/collector config), which exercises the regex compilation.

## Steps to reproduce

1. Configure an OTLP output with, e.g.:
   ```yaml
   outputs:
     vector:
       type: otlp
       endpoint: 127.0.0.1:4317
       protocol: grpc
       counter-patterns:
         - "counters"
         - "octets|packets|pkts|bytes"
   ```
2. Subscribe to a path containing a counter, e.g. `/interfaces/interface/state/counters/out-octets`.
3. Run via `gnmic subscribe` and inspect the exported OTLP metric type (e.g. with a Vector OTLP source + `vector tap`).

**Expected:** metrics whose value key matches a pattern are exported as monotonic cumulative `Sum`.
**Actual:** all metrics are exported as `Gauge`; `counter-patterns` is a silent no-op.

## Affected files

- `pkg/outputs/otlp_output/otlp_output.go` — `Init`, `setDefaultsFor`, `validateConfig`
- `pkg/outputs/otlp_output/otlp_converter.go` — `isCounter`, `convertEventToMetrics`

mvanhorn:
## Summary

The OTLP output now honors `counter-patterns` in subscribe mode, exporting matching metrics as counters (monotonic Sums) instead of gauges.

## Why this matters

As reported in #880, the compiled `counterRegexes` slice is populated only inside `validateConfig()`, which `Update()` and `Validate()` call but `Init()` never did. `Init()` (the standard subscribe path) called `setDefaultsFor()` and then stored the config directly, so the running config kept `counterRegexes == nil`, `isCounter()` always returned false, and every numeric metric took the Gauge branch. The option only worked in collector/dynamic mode where the API triggers `Update()`.

The fix calls `o.validateConfig(ncfg)` in `Init()` right after `setDefaultsFor()` and before storing the config, mirroring exactly what `Update()` already does (setDefaultsFor -> validateConfig -> store). This compiles `CounterPatterns` into the config that `Init()` actually stores, so `isCounter()` sees the patterns in subscribe mode. The error from `validateConfig` is returned so an invalid regex fails `Init()` cleanly, matching the Update path. No changes to the converter are needed since `isCounter()`/`convertEventToMetrics()` are already correct once the regexes are populated.

## Testing

Added three unit tests in `pkg/outputs/otlp_output/otlp_output_test.go`, all using an unreachable endpoint (no live device needed, following the existing `TestOTLP_InitSucceedsWithUnreachableEndpoint` pattern):

- valid counter-patterns compile into a non-nil `counterRegexes` of the right length and `isCounter()` returns true for a matching key, false for a non-matching one
- empty counter-patterns leave `counterRegexes` non-nil but empty (default gauge behavior preserved)
- an invalid regex makes `Init()` return an error

`go test ./pkg/outputs/otlp_output/...`, `go vet ./pkg/outputs/otlp_output/...`, and `gofmt -l` all pass.

Fixes #880

### DIFF
--- a/pkg/outputs/otlp_output/otlp_output.go
+++ b/pkg/outputs/otlp_output/otlp_output.go
@@ -197,6 +197,9 @@ func (o *otlpOutput) Init(ctx context.Context, name string, cfg map[string]inter
 		ncfg.Name = options.Name
 	}
 	o.setDefaultsFor(ncfg)
+	if err := o.validateConfig(ncfg); err != nil {
+		return err
+	}
 
 	// Apply logger
 	o.logger = outputs.BindLogger(options.Logger, outputType, ncfg.Name)
--- a/pkg/outputs/otlp_output/otlp_output_test.go
+++ b/pkg/outputs/otlp_output/otlp_output_test.go
@@ -1196,6 +1196,76 @@ func TestOTLP_ChannelCloseFlushes(t *testing.T) {
 	assert.Greater(t, server.ReceivedMetricsCount(), 0, "Remaining batch should be flushed when channel closes")
 }
 
+// TestOTLP_InitCompilesCounterPatterns verifies that Init (the subscribe path)
+// compiles counter-patterns into counterRegexes so that isCounter honors them at
+// runtime. Before the fix, Init only called setDefaultsFor and stored the config
+// without validateConfig, leaving counterRegexes nil so every metric exported as a
+// Gauge in subscribe mode.
+func TestOTLP_InitCompilesCounterPatterns(t *testing.T) {
+	cfg := map[string]interface{}{
+		"endpoint":         "unreachable-host:4317",
+		"protocol":         "grpc",
+		"timeout":          "1s",
+		"counter-patterns": []string{"counters", "octets|packets"},
+	}
+
+	output := &otlpOutput{}
+	err := output.Init(context.Background(), "test-otlp", cfg,
+		outputs.WithConfigStore(gomap.NewMemStore(store.StoreOptions[any]{})),
+	)
+	require.NoError(t, err, "Init should succeed with valid counter-patterns")
+	defer output.Close()
+
+	stored := output.cfg.Load()
+	require.NotNil(t, stored)
+	require.NotNil(t, stored.counterRegexes, "counterRegexes should be compiled by Init")
+	assert.Len(t, stored.counterRegexes, 2)
+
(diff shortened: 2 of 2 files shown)


karimra (approved):

karimra:
Thanks for the contribution!

mvanhorn:
Appreciate it, @karimra - counter-pattern validation should catch bad subscribe configs early now.

mvanhorn:
Appreciate the merge, @karimra! Good to have counter-pattern validation in OTLP subscribe mode.

PR #889, opened 2026-07-06
feat(otlp_output): add OTLP/HTTP transport and harden output lifecycle

drewelliott:
## Summary

Adds an OTLP/HTTP transport to the `otlp` output, complementing the existing gRPC path, plus lifecycle and correctness hardening that applies to both transports.

### OTLP/HTTP transport

- New `protocol: http` (default remains `grpc`). Endpoints accept either a bare `host:port` (port required, scheme derived from whether `tls` is configured, path defaults to `/v1/metrics`) or a full URL (scheme and path preserved; empty or `/` path defaults to `/v1/metrics`).
- mTLS via the existing `tls` block. `http://` combined with a `tls` block is rejected at Init to prevent a silent mTLS bypass.
- Per-request body compression (`compression: gzip|none`), default `gzip`; ignored by gRPC. `Content-Type`/`Content-Encoding` are protocol-owned and cannot be set or leaked through user headers in either direction.
- Error classification per the OTLP spec: `429/502/503/504` retryable, other `4xx/5xx` permanent, transport-level errors retryable. `Retry-After` honored up to a 30s cap; otherwise exponential backoff with full jitter (`math/rand/v2`).
- HTTP redirects are never followed (`ErrUseLastResponse`); a redirect classifies as a permanent error. Go forwards user-defined headers cross-host and 307/308 replay the body, so following redirects would allow a redirecting collector to receive the payload and headers at another origin.
- `PartialSuccess` (both transports) is never retried — retrying would duplicate the accepted points. Accounting is request-level: the batch counts as sent; rejected points are counted by the new `gnmic_otlp_output_rejected_data_points_total`, which is the data-loss signal to alert on.

### Lifecycle hardening (both transports)

- Live reload can no longer tear the config/transport pair mid-batch: an `atomic.Pointer` bundles `(cfg, transport)`, and transport teardown waits on a per-transport in-flight WaitGroup instead of wall-clock heuristics — including across config-only reloads layered on the same transport.
- `Update` publishes nothing until all fallible work has succeeded, so a failed reload leaves the previous state fully intact. Header-only reloads take effect on the next batch without a transport rebuild.
- `Close` no longer closes the event channel (producers dispatched by the outputs manager are not synchronized with `Close`; a send on a closed channel would panic the process). Workers exit via context cancellation; `Write`/`WriteEvent` tolerate post-`Close` state.
- Numeric config options are validated at Init/reload; previously a negative `batch-size`/`interval`/`buffer-size`/`num-workers` panicked at runtime (in a worker goroutine for the first two, killing the process).
- Bare endpoints on both transports must carry an explicit port (`net.SplitHostPort`), failing at Init instead of on first dial; gRPC schemed forms (`dns:///`, `unix:///`, `xds:///`) pass through unchanged.
- Config-reload logging redacts header values (`Authorization` bearer tokens etc.); key names stay visible.
- Prometheus collectors are tracked per instance: a failed Init rolls back only its own registrations, and Close unregisters them so a later Init on the same registry succeeds.

### Docs

`docs/user_guide/outputs/otlp_output.md` rewritten example-first (gRPC, HTTP, HTTP+mTLS), with a full configuration reference and sections for transport behavior, TLS/compression, metric naming, resource attributes, headers, retries/partial-success accounting.

### Tests

New coverage: end-to-end OTLP/HTTP delivery against a real mTLS server; endpoint validation (scheme/userinfo/port/TLS-coherence); Retry-After parsing and backoff capping; PartialSuccess on both transports including the accounting semantics; gzip compression and Content-Encoding ownership (including the `compression: none` strip); redirect refusal; header redaction; numeric config validation; race regressions for worker restart, two-step reload transport sharing, concurrent `WriteEvent`-vs-`Close`, and reload-path coverage through real `Update()` calls. The package passes `go test -race`.

### Known follow-ups (pre-existing, observed while developing — happy to file issues)

- The outputs manager never passes `WithRegistry` to outputs, so per-output `enable-metrics` has no scrapeable registry when driven through the manager.
- The manager dispatches `Write` goroutines unsynchronized with `Close` for all output types; this PR makes the `otlp` output itself safe against that, but the contract may be worth tightening centrally.
- The shared TLS helper does not set an explicit `MinVersion` (currently inherits Go's 1.2 default).
- A `buffer-size` reload swaps the event channel and can drop events still buffered in the old one (documented).

### DIFF
--- a/pkg/outputs/otlp_output/otlp_http.go
+++ b/pkg/outputs/otlp_output/otlp_http.go
@@ -267,6 +267,11 @@ func (o *otlpOutput) sendHTTP(ctx context.Context, state *outputState, req *metr
 	}
 	cfg := state.cfg
 
+	if err := o.validateRequest(req); err != nil {
+		o.logger.Error("validation error", "err", err)
+		return fmt.Errorf("request validation failed: %w", err)
+	}
+
 	raw, err := proto.Marshal(req)
 	if err != nil {
 		// proto.Marshal can fail for requests containing invalid UTF-8 in string
@@ -339,8 +344,13 @@ func (o *otlpOutput) sendHTTP(ctx context.Context, state *outputState, req *metr
 		}
 		var respProto metricsv1.ExportMetricsServiceResponse
 		if err := proto.Unmarshal(respBody, &respProto); err != nil {
-			// Malformed response body but transport-level success — log and accept.
-			o.logger.Debug("failed to unmarshal OTLP response body", "err", err)
+			// Malformed response body but transport-level success — accept (retrying
+			// would duplicate data) but surface it: a collector answering 2xx with a
+			// non-OTLP body is misconfigured or broken.
+			o.logger.Warn("failed to unmarshal OTLP response body", "err", err)
+			if cfg.EnableMetrics {
+				otlpMalformedResponses.WithLabelValues(cfg.Name).Inc()
+			}
 			return nil
 		}
 		if respProto.PartialSuccess != nil && respProto.PartialSuccess.RejectedDataPoints > 0 {
--- a/pkg/outputs/otlp_output/otlp_http.go
+++ b/pkg/outputs/otlp_output/otlp_http.go
@@ -1,0 +1,418 @@
+// © 2026 NVIDIA Corporation
+//
+// This code is a Contribution to the gNMIc project ("Work") made under the Google Software Grant and Corporate Contributor License Agreement ("CLA") and governed by the Apache License 2.0.
+// No other rights or licenses in or to any of NVIDIA's intellectual property are granted for any other purpose.
+// This code is provided on an "as is" basis without any warranties of any kind.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package otlp_output
+
+import (
+	"bytes"
+	"compress/gzip"
+	"context"
+	"crypto/tls"
+	"errors"
+	"fmt"
+	"io"
+	rand "math/rand/v2"
+	"net"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+	"time"
+
+	metricsv1 "go.opentelemetry.io/proto/otlp/collector/metrics/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+const otlpHTTPMetricsPath = "/v1/metrics"
+
+// httpExportError carries the HTTP status code alongside the underlying error,
+// so the caller can distinguish retryable (429/5xx) from permanent (4xx) failures.
+type httpExportError struct {
+	status     int
+	retryAfter time.Duration // populated from Retry-After header if present; 0 otherwise
+	msg        string
+}
+
+func (e *httpExportError) Error() string {
+	return fmt.Sprintf("HTTP export returned status %d: %s", e.status, e.msg)
+}
+
+// isPermanentHTTPError returns true when the error should NOT be retried.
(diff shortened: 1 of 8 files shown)


drewelliott:
I have been running this build fleet-wide for ~month with no issues.

drewelliott:
Rebased onto current main to resolve conflicts with #887. One thing surfaced during the rebase: main currently carries two identically-named `TestOTLP_InitCompilesCounterPatterns` functions in `otlp_output_test.go` (an artifact of the #886/#887 merge), which prevents the package's tests from compiling. This PR resolves that by renaming the conversion-focused one to `TestOTLP_InitCounterPatternsClassifyAsSum` and keeps both tests' coverage.

pkg/outputs/otlp_output/otlp_http.go
```diff
@@ -0,0 +1,418 @@
...
+		respBody, _ := io.ReadAll(io.LimitReader(resp.Body, 64*1024))
+		if len(respBody) == 0 {
+			return nil
+		}
+		var respProto metricsv1.ExportMetricsServiceResponse
+		if err := proto.Unmarshal(respBody, &respProto); err != nil {
+			// Malformed response body but transport-level success — log and accept.
+			o.logger.Debug("failed to unmarshal OTLP response body", "err", err)
+			return nil
+		}
```
karimra, line 355: Worth surfacing this error as a Warning log entry or a failure metric (or both).

pkg/outputs/otlp_output/otlp_http.go
```diff
@@ -0,0 +1,418 @@
...
+	resolvedURL, err := resolveMetricsURL(endpoint, cfg.TLS != nil)
+	if err != nil {
+		return nil, err
+	}
+
+	o.logger.Info("initialized OTLP HTTP client", "endpoint", resolvedURL)
+	return &httpClientState{client: client, endpoint: resolvedURL}, nil
+}
+
+func (o *otlpOutput) sendHTTP(ctx context.Context, state *outputState, req *metricsv1.ExportMetricsServiceRequest) error {
```
karimra, line 259: sendHTTP() should call validateRequest(req) just like sendGRPC()

karimra (changes requested):

pkg/outputs/otlp_output/otlp_http.go
drewelliott: I rebased the fix commit on top (a40e001) - I decided to do both, the unmarshal failure is now logged at Warn and when enable-metrics is set, counted by a new gnmic_otlp_output_malformed_responses_total counter. 

It's registered alongside the existing output metrics and documented in the metrics table. Kept the return nil since the export was delivered at the transport level and retrying would duplicate data.

pkg/outputs/otlp_output/otlp_http.go
drewelliott: I rebased the fix commit on top (a40e0013) - sendHTTP now calls validateRequest before marshaling, with the same logging and error wrapping as sendGRPC. 

The existing sendHTTP tests were sending empty ExportMetricsServiceRequest{} payloads, which now (correctly) fail validation, so they were updated to use a minimal structurally-valid request and a new test verifies an invalid request is rejected before anything goes on the wire.

karimra (approved):

karimra:
Thanks for the contribution and for adding the requested changes.

Issue #529, opened 2024-10-10
gNMIc REST API  interface protection


vseregin63:
Hi, Team!

We use gnmic cluster of two nodes. In such a cluster master node assign targets to slave via http rest api. Is where is a way to protect rest api interfaces of nodes from another network connection? May be user/pass auth, token auth or tls certs auth are supported on rest api? Please direct to docs.

Thnx!

vseregin63:
Have a good day!

I found in docs how to configure api server with tls auth required (https://gnmic.openconfig.net/user_guide/api/api_intro/) and added rootCA and server key/cert to gnmic config file

api-server:
  tls:
    ca-file: /var/run/secrets/rootCA.crt
    cert-file: /var/run/secrets/gnmic_api_server.pem
    key-file: /var/run/secrets/gnmic_api_server.key
    client-auth: "require-verify"

It works, but i still do not understand where to put client cert/key bacuse it needed for intra cluster nodes interconnection. Without it leader tries to assign element to slave gnmic via protected API and fails (doesn`t have a cert)

2024/10/22 12:36:38.240084 [gnmic] failed assigning target "leaf1" to service "gnmic-api": Post "https://server.com:7890/api/v1/config/targets": remote error: tls: certificate required

karimra:
mTLS is currently not supported by the clustering client. I agree that it will be useful, I will look into it.

vseregin63:
Thanks for reply, @karimra! If you could implement this, it will be great!

vseregin63:
@karimra , we checked mTLS interaction between cluster nodes and confirm it works ! Thank for your help, closing the issue

PR #892, opened 2026-07-15
fix(collector): honor clustering.tls for leader→member API calls

vuthinh0301:
# PR title

fix(collector): honor clustering.tls for leader→member API calls

# PR body

## Problem

In **collector mode** (`gnmic collect`), the cluster leader's HTTP clients that call other members' REST APIs (target assignment / unassignment / deletion) were plain `&http.Client{}` instances that never loaded the `clustering.tls` configuration — the client in `pkg/collector/managers/cluster/cluster_manager.go` even carried a literal `// TODO:`.

As a result, a TLS-secured API broke target dispatch:
- **private-CA server cert** on `api-server.tls` → `x509: certificate signed by unknown authority`
- **mTLS** (`api-server.tls.client-auth: require-verify`) → the dispatch client presents no client cert, so the leader gets `remote error: tls: bad certificate` calling members, **including itself** → no target ever assigned, no telemetry.

This works in **subscribe mode**, where `clustering.tls` was wired up in #544 (v0.39.0, for #529). Collector mode (v0.43.0) has a separate cluster implementation that never got the same treatment. The config was already parsed and env-expanded (`pkg/config/clustering.go`, `pkg/collector/env/env.go:18-21`) but never consumed anywhere under `pkg/collector/`.

## Fix

Add `newClusteringHTTPClient(*config.Clustering)` which builds the client from `clustering.tls` via `utils.NewTLSConfig(...)`, mirroring `(*App).createAPIClient` in `pkg/app/clustering.go`. Inject it into both the `ClusterManager.apiClient` (replacing the `// TODO:` client) and the assigner (`NewAssigner` now takes the client).

When `clustering.tls` is unset the behavior matches subscribe mode (skip-verify fallback), so plain-HTTP and public-CA setups are unaffected.

## Testing

- `go build ./...` — OK (go 1.25)
- `go vet ./pkg/collector/managers/cluster/...` — clean
- `go test ./pkg/collector/managers/cluster/... ./pkg/app/...` — pass
- Manually verified end-to-end on a 3-instance `gnmic collect` cluster + Consul + Nokia SR Linux target: before the fix, enabling `api-server.tls` with a private CA (mTLS) produced the `x509`/`bad certificate` errors above and no target was assigned; after the fix, with matching `clustering.tls`, the leader dispatches over HTTPS+mTLS, the lock is acquired, and telemetry flows. Subscribe mode with the same config already worked and still does.

## Files

- `pkg/collector/managers/cluster/cluster_manager.go`
- `pkg/collector/managers/cluster/assigner.go`
- `pkg/collector/managers/cluster/cluster_integration_test.go`

### DIFF
--- a/pkg/collector/managers/cluster/assigner.go
+++ b/pkg/collector/managers/cluster/assigner.go
@@ -41,13 +41,19 @@ type restAssigner struct {
 	logger *slog.Logger
 }
 
-func NewAssigner(store *collstore.Store) Assigner {
+// NewAssigner returns an Assigner that pushes assignments to cluster members.
+// client is the HTTP client built from clustering.tls (see
+// newClusteringHTTPClient); if nil, a plain client is used as a fallback.
+func NewAssigner(store *collstore.Store, client *http.Client) Assigner {
+	if client == nil {
+		client = &http.Client{
+			Timeout: 10 * time.Second,
+		}
+	}
 	return &restAssigner{
 		store:  store,
 		logger: slog.With("component", "assignment-pusher"),
-		client: &http.Client{
-			Timeout: 10 * time.Second,
-		},
+		client: client,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


karimra:
closing this in favor of #903. It solves the same issue but with slightly better behavior:
PR 903 clones `http.DefaultTransport` preserving proxy settings and has more comprehensive tests.

PR #893, opened 2026-07-17
fix(otlp_output): parse RFC 7951 numeric JSON strings as metric values

Mesverrum:
This change aligns the OTEL output with the way the Prometheus output works.

json_ietf encodes YANG uint64/int64/decimal64 as JSON strings. The OTLP converter previously treated every string as an enum (value=1 + value attribute), causing wrong Prom values and label cardinality explosion. Prefer numeric datapoints when the string parses, matching prometheus_output.


karimra (changes requested):
Thanks for this. The direction is correct but it needs a small change:
The change bypasses `strings-as-attributes`. The default for is `false`, that means string values are dropped. With this change, any string that parses as a number becomes a data point, even when the option is false. So users on the default config will see new metrics that did not exist before.

I think the numeric check should run after the `strings-as-attributes` gate, or use its own option.

PR #895, opened 2026-07-23
Stop retrying terminal subscription errors

dlanov:
Fixes #834.

## Summary

* classify gRPC subscription errors as retryable or terminal
* stop retrying terminal errors across `Subscribe`, `SubscribeChan`, and `SubscribeStreamChan`
* preserve retries for transient and unclassified errors for compatibility with targets that do not return canonical gRPC statuses
* distinguish locally initiated cancellation from a remotely returned `CANCELED` status
* add table-driven and behavioral tests for subscription retry handling

## Retry behavior

The following canonical gRPC statuses stop the subscription retry loop:

* `CANCELED`
* `INVALID_ARGUMENT`
* `NOT_FOUND`
* `ALREADY_EXISTS`
* `PERMISSION_DENIED`
* `RESOURCE_EXHAUSTED`
* `FAILED_PRECONDITION`
* `OUT_OF_RANGE`
* `UNIMPLEMENTED`
* `DATA_LOSS`
* `UNAUTHENTICATED`

`RESOURCE_EXHAUSTED` is treated as terminal for subscriptions as requested in #834.

Transient statuses such as `DEADLINE_EXCEEDED`, `ABORTED`, and `UNAVAILABLE` remain retryable. `UNKNOWN`, `INTERNAL`, and non-gRPC errors also retain the existing retry behavior for compatibility with older targets and proxies.

Locally initiated cancellation remains silent, while a remote `CANCELED` status is reported once before the retry loop stops.

## Testing

* `go test ./target -count=1 -race`
* `go test ./... -count=1`
* `go vet ./...`
* `CGO_ENABLED=0 ./tests/run_tests.sh`

### DIFF
--- a/pkg/api/target/subscribe.go
+++ b/pkg/api/target/subscribe.go
@@ -49,10 +49,18 @@ SUBSC_NODELAY:
 		nctx = t.appendRequestMetadata(nctx)
 		subscribeClient, err = t.Client.Subscribe(nctx, t.callOpts()...)
 		if err != nil {
-			t.errors <- &TargetError{
-				SubscriptionName: subscriptionName,
-				Err:              fmt.Errorf("failed to create a subscribe client, target='%s', retry in %s. err=%v", t.Config.Name, t.Config.RetryTimer, err),
+			if isLocalCancellation(nctx, err) {
+				cancel()
+				return
 			}
+			if !isRetryableSubscribeError(err) {
+				sendError(t.errors, ctx, subscriptionName,
+					fmt.Errorf("failed to create a subscribe client, target='%s': %w", t.Config.Name, err))
+				cancel()
+				return
+			}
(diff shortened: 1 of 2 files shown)


XioNoX:
For `INVALID_ARGUMENT` we should keep retrying (or have it configurable).

On Juniper MX, the path `/system/alarms/alarm/state/id` doesn't exist if there are no active alarms, so the device returns a `InvalidArgument`:

`time=2026-09-02T09:25:46.614Z level=ERROR msg="subscription error" component=targets-manager error="&{SubscriptionName:alarms Err:rpc error: code = InvalidArgument desc = Unsupported subscription path, /system/alarms/alarm/state/id}"`

But it show up as soon as an alarm is active. So if we stop retrying we would never be notified of an alarm.

PR #898, opened 2026-07-25
feat(output): add custom headers support to kafka

senthilsam:
Added support for adding custom headers to Kafka messages

```json
kafka:
  add-headers:
      env: prod
      sub: '{{ index .Meta "subscription-name" }}'
      source: '{{ index .Meta "source" }}'
```

### DIFF
--- a/pkg/outputs/kafka_output/kafka_output.go
+++ b/pkg/outputs/kafka_output/kafka_output.go
@@ -9,6 +9,7 @@
 package kafka_output
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -92,10 +93,12 @@ type kafkaOutput struct {
 }
 
 type dynConfig struct {
-	targetTpl *template.Template
-	msgTpl    *template.Template
-	evps      []formatters.EventProcessor
-	mo        *formatters.MarshalOptions
+	targetTpl  *template.Template
+	msgTpl     *template.Template
+	evps       []formatters.EventProcessor
+	mo         *formatters.MarshalOptions
+	headers    []sarama.RecordHeader
+	headerTpls map[string]*template.Template
 }
 
 // config //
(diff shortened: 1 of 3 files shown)


pkg/outputs/kafka_output/kafka_output.go
```diff
@@ -688,10 +704,25 @@ CRPROD:
...
+					headers = append(headers, sarama.RecordHeader{
+						Key:   []byte(k),
+						Value: []byte(v),
+					})
+				}
+
+				headers = append(headers, sarama.RecordHeader{
+					Key:   []byte("sub"),
+					Value: []byte(m.GetMeta()["subscription-name"]),
+				})
```
karimra: This is added unconditionally, either remove it or gate it behind a config knob.

pkg/outputs/kafka_output/kafka_output.go
```diff
@@ -688,10 +704,25 @@ CRPROD:
...
+
 				topic := k.selectTopic(m.GetMeta())
 				msg := &sarama.ProducerMessage{
-					Topic: topic,
-					Value: sarama.ByteEncoder(b),
+					Topic:     topic,
+					Value:     sarama.ByteEncoder(b),
+					Headers:   headers,
+					Timestamp: time.Now(),
```
karimra: Timestamp is not required, it is broker assigned and unrelated to the stated PR scope.
Create another PR to add it if required (behind a config knob)

pkg/outputs/kafka_output/kafka_output.go
```diff
@@ -688,10 +704,25 @@ CRPROD:
 					}
 				}
 
+				var headers []sarama.RecordHeader
+				for k, v := range cfg.AddHeaders {
+					headers = append(headers, sarama.RecordHeader{
+						Key:   []byte(k),
+						Value: []byte(v),
+					})
+				}
```
karimra: Duplicated code in sync and async producers. Good to create a helper and call in both locations.
Static headers don't need to be computed for every message. Compute once, store as part of `dynConfig`.

karimra (changes requested):
please add docs for the added config knob

pkg/outputs/kafka_output/kafka_output.go
senthilsam: Taken, introduced template for adding dynamic tags

pkg/outputs/kafka_output/kafka_output.go
senthilsam: addressed

pkg/outputs/kafka_output/kafka_output.go
senthilsam: removed i will open a separate pr for this

pkg/outputs/kafka_output/kafka_output.go
```diff
@@ -160,6 +163,70 @@ func (k *kafkaOutput) buildEventProcessors(logger *slog.Logger, eventProcessors
...
+func buildKafkaHeaders(headers map[string]string) ([]sarama.RecordHeader, map[string]*template.Template, error) {
+	if len(headers) == 0 {
+		return nil, nil, nil
+	}
+
+	staticHeaders := make([]sarama.RecordHeader, 0, len(headers))
+	dynamicHeaders := make(map[string]*template.Template)
+
+	for k, v := range headers {
+		if strings.Contains(v, "{{") {
```
karimra: There is no need to distinguish between Go templates and static strings. A static string renders as itself if used as a Go template. Just treat all values as Go templates and render them all in the same way.

karimra (changes requested):

karimra (approved):

PR #902, opened 2026-07-27
config: apply environment overrides before unmarshal

dlanov:
Fixes #833

Nested free-form collector configuration supplied through `GNMIC_*`
environment variables was not reflected in the unmarshaled `Config`.

Apply environment-backed free-form configuration before unmarshaling,
while limiting underscore-based nesting to sections whose keys are not
known ahead of time:

- `clustering`
- `outputs`
- `inputs`
- `processors`
- `loader`
- `actions`

Known struct-backed configuration such as `cluster-name`, `log-file`,
and `format` continues to use Viper's `AutomaticEnv` and environment
key replacer. This avoids treating underscores in variables such as
`GNMIC_LOG_FILE` as nesting separators while still supporting nested
values such as `GNMIC_CLUSTERING_LOCKER_ADDRESS`.

Regression coverage verifies:

- known struct-backed environment variables remain handled by Viper
- `GNMIC_LOG_FILE` and `GNMIC_CLUSTER_NAME` override their fields
- `GNMIC_CLUSTERING_LOCKER_ADDRESS` overrides nested free-form config
- unrelated clustering values from the configuration file are preserved
- underscore splitting is restricted to the six free-form sections

Testing:

- focused environment/config regression tests
- `go test ./pkg/config -count=1`
- `go test ./... -count=1`
- `go vet ./...`
- `git diff --check`

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -361,12 +361,13 @@ func (c *Config) Load(ctx context.Context) error {
 		}
 	}
 
+	c.mergeEnvVars()
+
 	err := c.FileConfig.Unmarshal(c)
 	if err != nil {
 		return err
 	}
 
-	c.mergeEnvVars()
 	return c.expandOSPathFlagValues()
 }
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -359,12 +359,13 @@ func (c *Config) Load(ctx context.Context) error {
 		}
 	}
 
+	c.mergeEnvVars()
+
 	err := c.FileConfig.Unmarshal(c)
 	if err != nil {
 		return err
 	}
 
-	c.mergeEnvVars()
 	return c.expandOSPathFlagValues()
 }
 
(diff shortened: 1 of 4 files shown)


karimra (changes requested):
The real problem is that _ has two meanings. In log-file it is a dash, and log-file is a struct field. In clustering.locker.address it is a level of nesting, and that key lives inside a map[string]any. envToMap cannot know which meaning to use. Moving the call does not fix this. It only changes a silent failure into a crash at startup.

Could you apply the _ splitting only to the free-form sections (clustering, outputs, inputs, processors, loader, actions)? For the other keys, viper's AutomaticEnv and the key replacer already work, because those keys are struct fields with known names. This way the guessing happens only where the keys are not known in advance.

dlanov:
Thanks for the clarification. I updated the implementation so underscore splitting is limited to the free-form sections: clustering, outputs, inputs, processors, loader, and actions. Known struct-backed fields continue through Viper AutomaticEnv/key replacement. I also added regression coverage for GNMIC_LOG_FILE and GNMIC_CLUSTER_NAME alongside GNMIC_CLUSTERING_LOCKER_ADDRESS, and updated the existing environment helper tests to reflect the scoped behavior. Full go test ./..., go vet ./..., and git diff --check pass locally.

karimra (approved):

karimra:
Thanks for the contribution!

PR #905, opened 2026-08-03
fix(tcp): retry messages after write failures

### LINKED ISSUE #904: The first telemetry message is dropped when the output type is TCP
gnmic version 0.46.0
configuration:

listen: true
address: "0.0.0.0:57400"
insecure: true
outputs:
  to-collector:
    type: tcp
    address: collector:9100
    format: event
    delimiter: "\n"

When TCP session to collector does not exist, and gnmic receives telemetry data, it opens TCP connection with collector but does not forward received data to collector. All subsequent telemetry data is forwarded normally if TCP session with collector is established.

dlanov:
## Summary

The TCP output worker currently removes a telemetry message from its buffer before writing it to the active connection. If that write fails, the worker reconnects but does not retain the dequeued message, causing the first message at the connection failure boundary to be dropped.

This change:

- keeps the current framed payload pending across transient write/reconnect failures;
- reconnects and retries the pending payload before dequeuing another message;
- adds configurable `max-retries` for pending messages, defaulting to 3;
- drops a pending message after its retry budget is exhausted so one unwritable message cannot block the worker forever;
- exposes TCP output delivery-error and dropped-message Prometheus metrics when `enable-metrics` is enabled;
- supports enabling TCP metrics through a live output update;
- preserves per-worker message ordering across reconnects;
- handles short writes by continuing until the complete payload is written;
- makes TCP dialing and retry delays responsive to context cancellation; and
- closes failed connections before reconnecting.

Delivery across an ambiguous TCP write failure is at-least-once. A complete payload may be retried if part of it was accepted by the failed connection. After `max-retries` is exhausted, the pending message is dropped and the worker continues with the next buffered message.

## Metrics

When `enable-metrics: true`, the TCP output exposes:

- `gnmic_tcp_output_errors_total{name,reason}` with `reason` values `dial` or `write`;
- `gnmic_tcp_output_dropped_messages_total{name,reason}` for messages dropped after retry exhaustion.

## Testing

- `go test ./pkg/outputs/tcp_output -count=20`
- `go test -race ./pkg/outputs/tcp_output -run 'TestTCP_(RetriesMessageAfterWriteFailure|DropsMessageAfterRetryLimit|UpdateEnablesMetrics)$' -count=5`
- `go vet ./...`
- `staticcheck ./pkg/outputs/tcp_output`

Fixes #904

### DIFF
--- a/docs/user_guide/outputs/tcp_output.md
+++ b/docs/user_guide/outputs/tcp_output.md
@@ -44,10 +44,20 @@ outputs:
     keep-alive: 
     # time duration to wait before re-dial in case there is a failure
     retry-interval: 
-    # NOT IMPLEMENTED boolean, enables the collection and export (via prometheus) of output specific metricss
+    # maximum number of retries for a pending message after its initial failed delivery attempt.
+    # defaults to 3. Once exhausted, the message is dropped so the worker can continue.
+    max-retries: 3
+    # boolean, enables the collection and export (via prometheus) of output-specific metrics
     enable-metrics: false 
     # list of processors to apply on the message before writing
     event-processors: 
 ```
+
+When `enable-metrics` is set to `true`, the TCP output exposes:
+
+- `gnmic_tcp_output_errors_total{name,reason}` for delivery errors. The `reason` label is `dial` or `write`.
+- `gnmic_tcp_output_dropped_messages_total{name,reason}` for messages dropped after the retry budget is exhausted.
+
+`max-retries` applies only after a message has been dequeued for delivery. Dial failures while a message is pending count against its retry budget.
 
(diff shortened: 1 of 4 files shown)


karimra:
Good catch.
One question: An unwritable message will be retried forever, it's better if the message is dropped after N (configurable) attempts and the TCP output surfaces an error as a metric.

dlanov:
@karimra Thanks for pointing that out. I updated #905 so a pending TCP message no longer retries forever: `max-retries` is configurable and defaults to 3, and the worker drops the pending message after the retry budget is exhausted so it can continue with subsequent messages.

I also added TCP output Prometheus counters for dial/write delivery errors and messages dropped after retry exhaustion, including support for enabling the metrics through a live output update. The new bounded-retry and metrics paths are covered by focused and race tests.

karimra (approved):

karimra:
Thanks for the contribution and applying the requested changes!

PR #912, opened 2026-08-10
fix(influxdb_output): use-after-close on client rebuild

karimra:
Update() closed the old influx client immediately after swapping the pointer, while workers still held it and kept calling client.WriteAPI() Workers now run under a per-generation context and WaitGroup, so a rebuild starts the new generation, drains the old one, and only then closes the client it held.


hellt (approved):
LGTM

PR #913, opened 2026-08-10
fix(clickhouse_output): data race on test hook globals

karimra:
Init now snapshots the hooks into the output instance, so background goroutines never read the package vars a later test reassigns.


hellt (approved):
LGTM

PR #921, opened 2026-08-10
feat(logging): add --log-format option for JSON application logs

igorfernandes:
## Summary

Adds a `--log-format` option so application logs can be emitted as structured JSON in addition to the existing text format.

Today gNMIc always uses `slog.NewTextHandler`. This change keeps that as the default and allows selecting JSON via:

- CLI: `--log-format text|json` (default `text`)
- config file: `log-format: json`
- env: `GNMIC_LOG_FORMAT=json`

JSON logs use the standard `slog.JSONHandler` fields (`time`, `level`, `msg`, plus any attributes). The change is wired through both logging entry points (`InitAppLogging` and `logging.NewLogger`) so CLI and collector paths stay consistent.

## Testing

- `go test ./pkg/logging/ ./pkg/config/ -count=1 -run 'TestNewHandlerJSON|TestNormalizeLogFormat|TestLoggingFlags'`
- Smoke test:
  ```bash
  gnmic version --log --log-format json
  ```
  Example output:
  ```json
  {"time":"2026-08-07T10:08:05.153510462Z","level":"INFO","msg":"gnmic version","version":"dev","commit":"none","date":"unknown","gitURL":"","docs":"https://gnmic.openconfig.net"}
  ```


karimra (approved):
Thanks for the contribution!

Issue #899, opened 2026-07-25
Deleting target is not cleaning up subscription metrics


senthilsam:
When the target is removed via the loader, the subscribe metrics remain in the HTTP endpoint until restart

```
gnmic_subscribe_number_of_received_subscribe_response_messages_total
gnmic_subscribe_number_of_failed_subscribe_request_messages_total
```

senthilsam:
Fixed in #901

PR #931, opened 2026-08-16
fix(target): clean up target metrics on deletion

senthilsam:


senthilsam:
Fixes: https://github.com/openconfig/gnmic/issues/899

pkg/collector/managers/targets/targets_manager.go
```diff
@@ -647,6 +647,7 @@ func (tm *TargetsManager) remove(name string) {
 		mt.readerCancel = nil
 		mt.Unlock()
 	}
+	tm.deleteTargetMetrics(mt)
```
karimra, line 650: `mt` can be nil here

karimra (changes requested):

PR #943, opened 2026-08-19
fix(tcp): avoid blocking on full output buffer

### LINKED ISSUE #942: TCP output blocks subscription processing when its buffer is full
## Description

The TCP output blocks the caller indefinitely when its bounded channel is full. This can stop all subscription processing for a target, including unrelated outputs attached to the same subscription.

This is reproducible on `v0.47.0` (`747b576e8eab3b88a21d540e2a31c8d2183c1a74`) and current `main` (`597750387b5117f625a76681f6b4d83147f9151b`).

## Reproduction

1. Configure a target with a TCP output and at least one other output.
2. Set the TCP output `buffer-size` to `1`.
3. Make the configured TCP endpoint unavailable.
4. Send at least two subscription responses for the target.

The first response fills the TCP output buffer. The second call to `tcpOutput.Write` remains blocked at the channel send.

A regression test that invokes `Write` with an already-full buffer fails against current `main` with:

```text
--- FAIL: TestTCP_WriteDoesNotBlockWhenBufferIsFull (1.00s)
    tcp_output_test.go: full TCP output buffer blocked the caller
```

## Actual behavior

- `tcpOutput.Write` performs an unconditional channel send after checking the context only once: [`(*buffer) <- b`](https://github.com/openconfig/gnmic/blob/747b576e8eab3b88a21d540e2a31c8d2183c1a74/pkg/outputs/tcp_output/tcp_output.go#L368-L390).
- When the TCP endpoint is unavailable, workers retry dialing before dequeuing a buffered message, so the buffer is not drained: [`startWithDialer`](https://github.com/openconfig/gnmic/blob/747b576e8eab3b88a21d540e2a31c8d2183c1a74/pkg/outputs/tcp_output/tcp_output.go#L511-L548).
- The subscription listener waits for all configured outputs to return before reading the next response: [`export`](https://github.com/openconfig/gnmic/blob/747b576e8eab3b88a21d540e2a31c8d2183c1a74/pkg/app/subscribe.go#L294-L379).
- Canceling the subscription context after the send blocks does not release the goroutine.
- `max-retries` does not bound this condition because no message has been dequeued into `pending` yet.

## Expected behavior

A full TCP output buffer must not block the target's subscription processing indefinitely. New messages should be dropped when the bounded buffer is full, and the existing dropped-message metric should expose the loss.

## Proposed fix

- Make enqueue non-blocking and drop new messages when the TCP output buffer is full.
- Increment `gnmic_tcp_output_dropped_messages_total{name,reason="buffer_full"}`.
- Document the overflow behavior.
- Add a regression test at the `Write` boundary.

I have a tested patch ready and will link the pull request here.

pood1e:
## Summary

The TCP output currently performs a blocking send to its internal bounded channel. If the TCP endpoint is unavailable and the buffer fills, `Write` never returns. Because the subscription export path waits for every output, this stalls subsequent subscription processing for the target and can also hold the goroutine after context cancellation.

This change:

- makes TCP output enqueue non-blocking;
- drops new messages when the configured buffer is full;
- records each overflow through `gnmic_tcp_output_dropped_messages_total{name,reason="buffer_full"}` when metrics are enabled;
- documents the buffer overflow behavior and remediation; and
- adds a regression test at the `Write` boundary.

Messages already dequeued by a worker retain the existing `max-retries` delivery behavior. This only changes what happens after the bounded input buffer has exhausted its capacity.

## Testing

- `go test ./pkg/outputs/tcp_output -count=20`
- `go test -race ./pkg/outputs/tcp_output -run 'TestTCP_(WriteDoesNotBlockWhenBufferIsFull|RetriesMessageAfterWriteFailure|DropsMessageAfterRetryLimit|UpdateEnablesMetrics)$' -count=5`
- `go vet ./pkg/outputs/tcp_output`
- `go run honnef.co/go/tools/cmd/staticcheck@latest ./pkg/outputs/tcp_output`

Fixes #942

### DIFF
--- a/pkg/outputs/tcp_output/tcp_output.go
+++ b/pkg/outputs/tcp_output/tcp_output.go
@@ -386,13 +386,15 @@ func (t *tcpOutput) Write(ctx context.Context, m proto.Message, meta outputs.Met
 		}
 		buffer := t.buffer.Load()
 		for _, b := range bb {
-			t.enqueue(cfg, *buffer, b)
+			t.enqueue(ctx, cfg, *buffer, b)
 		}
 	}
 }
 
-func (t *tcpOutput) enqueue(cfg *config, buffer chan<- []byte, payload []byte) {
+func (t *tcpOutput) enqueue(ctx context.Context, cfg *config, buffer chan<- []byte, payload []byte) {
 	select {
+	case <-ctx.Done():
+		return
 	case buffer <- payload:
 	default:
 		t.logger.Warn("dropping TCP message because output buffer is full")
--- a/pkg/outputs/tcp_output/tcp_output.go
+++ b/pkg/outputs/tcp_output/tcp_output.go
@@ -386,7 +386,18 @@ func (t *tcpOutput) Write(ctx context.Context, m proto.Message, meta outputs.Met
 		}
 		buffer := t.buffer.Load()
 		for _, b := range bb {
-			(*buffer) <- b
+			t.enqueue(cfg, *buffer, b)
+		}
+	}
+}
+
+func (t *tcpOutput) enqueue(cfg *config, buffer chan<- []byte, payload []byte) {
+	select {
+	case buffer <- payload:
+	default:
+		t.logger.Warn("dropping TCP message because output buffer is full")
+		if cfg.EnableMetrics {
+			tcpOutputDroppedMessages.WithLabelValues(t.name, "buffer_full").Inc()
 		}
 	}
 }
--- a/docs/user_guide/outputs/tcp_output.md
+++ b/docs/user_guide/outputs/tcp_output.md
@@ -11,7 +11,8 @@ outputs:
     address: IPAddress:Port 
     # maximum sending rate, e.g: 1ns, 10ms
     rate: 10ms 
-    # number of messages to buffer in case of sending failure
+    # number of messages to buffer in case of sending failure.
+    # New messages are dropped when the buffer is full.
     buffer-size:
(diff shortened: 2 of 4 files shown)


pood1e:
@googlebot rescan

pkg/outputs/tcp_output/tcp_output.go
```diff
@@ -386,7 +386,18 @@ func (t *tcpOutput) Write(ctx context.Context, m proto.Message, meta outputs.Met
...
+	}
+}
+
+func (t *tcpOutput) enqueue(cfg *config, buffer chan<- []byte, payload []byte) {
+	select {
+	case buffer <- payload:
+	default:
+		t.logger.Warn("dropping TCP message because output buffer is full")
+		if cfg.EnableMetrics {
+			tcpOutputDroppedMessages.WithLabelValues(t.name, "buffer_full").Inc()
```
karimra, line 402: `enqueue` should take `ctx` as an arg and select on it so that it doesn't block when the context is canceled.

karimra (changes requested):

pkg/outputs/tcp_output/tcp_output.go
pood1e: Fixed in 6981707: `enqueue` now accepts the write context and selects on `ctx.Done()`. Added a regression test to ensure cancellation returns without recording a `buffer_full` drop.

karimra (approved):

karimra:
Thanks for the contribution!

PR #945, opened 2026-08-20
fix: release cluster locks during graceful shutdown

### LINKED ISSUE #944: SIGTERM does not stop the cluster locker, delaying target failover until lease expiry
### Description

The process signal handler cancels the application context and immediately calls `os.Exit(0)`, but it never calls `Locker.Stop()`. All built-in locker implementations use `Stop()` to release acquired locks and deregister the instance. Because `os.Exit` skips deferred cleanup, a Kubernetes rolling restart leaves the target and leader locks behind until their leases expire.

This makes a graceful `SIGTERM` behave like an ungraceful crash from the cluster's point of view.

### Reproduction

1. Run at least two gNMIc instances with clustering enabled and a Redis locker.
2. Assign one or more targets to an instance.
3. Send `SIGTERM` to the target owner.
4. Observe that the owner process exits, while its target locks remain in Redis until `lease-duration` expires.
5. The surviving member cannot acquire those targets before expiry.

### Expected behavior

On a graceful termination signal, gNMIc should first cancel its application context and then call the initialized locker's `Stop()` method before exiting. A locker failure should be logged and lease expiry should remain the fallback.

Unexpected crashes (`SIGKILL`, node loss) should continue to rely on lease expiry.

### Actual behavior

`setupCloseHandler` currently runs `CleanupPlugins()`, calls the context cancel function, and immediately exits. The initialized locker is never stopped.

### Impact

Planned Kubernetes rollouts unnecessarily introduce the full lease-duration interruption for affected targets. The drain API added by #579 is useful for explicit maintenance workflows, but normal Kubernetes `SIGTERM` handling should still release process-owned locks and service registration.

### Version

Reproduced against `main` at `597750387b5117f625a76681f6b4d83147f9151b`; the same lifecycle exists in v0.47.0.

pood1e:
## Summary

- centralize process shutdown in an idempotent `App.Shutdown` method
- cancel application work before stopping the configured locker
- release target/leader locks and service registration on `SIGTERM` instead of waiting for lease expiry
- prevent the command goroutine from exiting while signal-triggered cleanup is still running
- preserve lease expiry as the fallback when locker cleanup fails or the process is killed ungracefully

## Tests

- `go test ./...`
- `go test -race ./pkg/app ./pkg/cmd`

Fixes #944

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -122,6 +122,9 @@ type App struct {
 	// processors plugin manager
 	pm *plugin_manager.PluginManager
 
+	shutdownOnce sync.Once
+	shutdownErr  error
+
 	// pprof
 	pprof *pprofServer
 }
--- a/pkg/app/shutdown.go
+++ b/pkg/app/shutdown.go
@@ -1,0 +1,16 @@
+package app
+
+// Shutdown stops application work and releases resources that must not be left
+// to lease expiry. It is safe to call more than once.
+func (a *App) Shutdown() error {
+	a.shutdownOnce.Do(func() {
+		if a.Cfn != nil {
+			a.Cfn()
+		}
+		if a.locker != nil {
+			a.shutdownErr = a.locker.Stop()
+		}
+		a.CleanupPlugins()
+	})
+	return a.shutdownErr
+}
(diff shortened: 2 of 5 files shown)


karimra:
Is this still a draft, or is it ready for review ?

pood1e:
Ready for review now. I revalidated it against the current `main`: `go test ./...`, `go test -race ./pkg/app ./pkg/cmd`, and `go vet ./pkg/app ./pkg/cmd` all pass.

karimra (approved):

Issue #947, opened 2026-08-28
Prometheus Write backpressure retains per-target responses and can OOM subscribe mode


pood1e:
## Bug description

In `subscribe` mode, slow output transformation can retain up to `target count × target buffer size` complete gNMI `SubscribeResponse` protobufs. With many targets or large initial snapshots, memory grows until the process is OOM-killed.

The reproduced path is:

```text
gRPC Recv
  -> Target.handleStreamSubscriptionRcv
  -> per-target subscribeResponses channel
  -> App.StartTargetsManager
  -> App.export
  -> prometheus_write.Write
  -> unbuffered msgChan
  -> limited output workers
```

Each target's receive channel defaults to 100 responses. `App.export` waits for the selected outputs, while `prometheus_write.Write` waits for an available worker through an unbuffered channel. When event processing is slower than incoming initial snapshots, every target independently fills its receive buffer with complete protobuf payloads.

After the output timeout, `prometheus_write.Write` returns and drops the ingress message. That drop is not included in the Remote Write failure metric because no HTTP request was attempted.

This is not an unreachable-object leak. It is multiplicative, reachable retention caused by the absence of a process-wide bounded input queue and explicit overload accounting.

## Type

- Performance defect
- Reliability defect
- Design defect

## Reproduction

Tested in `subscribe` mode with `main` at `c820a545e8c32325e15c5a8cd595965f9e57ae32`:

1. Start 220 active target entries backed by 20 simulators.
2. Configure `prometheus_write` with 2 workers, 2 writers and a 10,000 time-series buffer.
3. Point Remote Write at a local endpoint that immediately returns HTTP 204, excluding network and remote storage latency.
4. Let all 220 streams reach `READY` and emit their initial snapshots.

Observed growth:

| Sample | Heap alloc | Heap sys | RSS | Goroutines | FDs |
| --- | ---: | ---: | ---: | ---: | ---: |
| Initial | 2.34 GB | 2.85 GB | 2.71 GB | 2019 | 228 |
| +44 seconds | 3.17 GB | 5.39 GB | 5.23 GB | 2018 | 228 |
| +2 minutes | 5.01 GB | 5.98 GB | 5.85 GB | 2018 | 228 |

[...]

PR #948, opened 2026-08-28
fix(prometheus_write): propagate cancellable input backpressure

pood1e:
Refs #947.

## Problem

In `subscribe` mode, `prometheus_write` previously admitted input through an
unbuffered channel with a timeout. Under sustained downstream pressure, the
timeout either lost telemetry or moved the backlog into per-target subscription
queues, where retained protobuf responses scaled with the target count.

## Changes

- add a bounded input queue configured by `input-buffer-size` (default `1`)
- propagate backpressure while preserving cancellation from the target lifecycle
  and output shutdown
- make downstream event and time-series handoffs cancellable as well
- reject runtime input queue resizing, which cannot be performed without loss or
  reordering
- expose input queue depth, capacity, backpressure count, and backpressure
  duration metrics
- document the input, target, and time-series buffer semantics

The queue does not silently drop messages. When it is full, the producer waits
until capacity becomes available, its context is canceled, or the output is
closed. A target sharing the output therefore slows down under sustained
pressure.

This bounds the output-local protobuf queue. Per-target subscription response
queues remain separately configured and default to `100`. Their buffer is
shared by every output assigned to the target, so this PR does not change that
global default. Deployments that require a smaller upstream burst buffer can set
the target `buffer-size` explicitly; that setting affects all outputs on the
target.

## Validation

- `go test -race ./...`
- `git diff --check`


pood1e:
Downstream 640-device validation showed that the current drop-on-full implementation bounds memory but does not preserve telemetry completeness: all three input queues remained saturated, `buffer_full` increased by 268,076 over 185 seconds, and representative interface metric coverage fell far below the expected fleet size.

I am marking this PR as draft while revising the design. The next revision will use a small immutable input queue with cancellation/output-lifecycle-aware blocking, so overload propagates to the target listener and gRPC HTTP/2 flow control instead of dropping admitted telemetry. Queue depth/capacity will remain observable, and wait count/duration will replace ingress-drop metrics.

The downstream deployment also reduces per-target response buffers and uses stable single-worker processing shards; those deployment changes are separate from the reusable upstream output fix.

pkg/config/targets.go
karimra, line 1: I don't think it's a good idea to change the default now. It's a global change that affects every output downstream of the target.

karimra (changes requested):

pood1e:
Addressed in `f277d74cf98d7a6aa011c7d9ac53ee9b717141e9`.

Agreed that the target buffer is shared upstream of output dispatch and should not change in a `prometheus_write`-specific PR. The default is restored to `100`, and the documentation now distinguishes the output-local input queue from the independently configured target queue.

Because the existing per-target backlog remains outside this PR, the PR now uses `Refs #947` instead of closing the issue. Deployments can explicitly lower target `buffer-size`, with the documented consequence that it affects every output assigned to that target. A shared admission/default change requires separate design review.

### Test validation

- Head commit: `f277d74cf98d7a6aa011c7d9ac53ee9b717141e9`
- `go test -race ./...` — PASS
- `git diff --check` — PASS
- GitHub checks: [`check-changes`](https://github.com/openconfig/gnmic/actions/runs/33849607444/job/100949156116) and Google CLA — PASS

pood1e:
Downstream validation found a measurement bug in the proposed backpressure metric: `defer histogram.Observe(time.Since(started).Seconds())` evaluates the elapsed duration when the defer is registered, before the write blocks. A long wait is therefore recorded as almost zero.

Type: observability correctness. Impact: deployments using this PR can underestimate input queue pressure; the blocking and cancellation behavior itself is unchanged.

I will defer a function that calculates elapsed time when `Write` actually returns, and extend the existing blocked-write test to verify the histogram duration. This is a small correction to the pending PR, with no new configuration or dependency.

pood1e:
The duration correction is in head `cc5ac7a73de939b824fcf8e9527caa54ee4720ee`.

The regression fills the one-entry input queue, keeps the second write blocked for 25 ms, then drains the queue and checks the histogram delta. Before the fix it reports `0.000000s` and fails the 20 ms lower bound. With the deferred closure, `go test -race ./pkg/outputs/prometheus_output/prometheus_write_output -count=1` passes, including queue pressure, cancellation and shutdown tests.

`gofmt` and `git diff --check` also pass. The test does not need a live telemetry target or a Remote Write server.

karimra (approved):

PR #951, opened 2026-08-28
feat(processor): add event field allow-list

### LINKED ISSUE #950: Wide events need a native field allow-list before expensive processors
## Bug description

`event-starlark` exposes `EventMsg.Values` as a mapping, but a Starlark program that iterates `event.values.items()` converts every value and allocates a tuple for every entry before the script can apply its field allow-list. Wide telemetry notifications therefore pay conversion and interpreter cost for fields that are immediately discarded.

This is especially expensive for database-backed telemetry. A production-shaped notification with 34,014 flattened values, sampled every 15 seconds, keeps each collector instance near three CPU cores and creates approximately 285 MB/s of allocations. A 30-second CPU profile attributes 76.84% cumulatively to Starlark function execution; an allocation profile attributes 88.06% cumulatively to the same path.

Existing processors do not provide the required operation:

- `event-allow` keeps or drops the whole event.
- `event-delete` removes matching fields, which requires enumerating every unwanted vendor field.
- Filtering in Starlark is too late because the full mapping is already traversed by the interpreter.

## Type

Performance / processor capability gap.

## Impact

Collectors processing wide notifications can accumulate minutes of telemetry lag while remaining healthy and returning successful Remote Write responses. Backpressure prevents unbounded memory growth, but it cannot compensate for transformation throughput below the input rate. The cost scales with the device's full table width rather than the small set of exported metrics.

## Proposed fix

Add a generic `event-keep` processor, complementary to `event-delete`, that removes value or tag entries which do not match configured regular expressions. Categories without configured selectors remain unchanged. The processor runs before expensive transformations and mutates the existing `EventMsg` maps in place.

Suggested configuration:

```yaml
processors:
  keep-exported-fields:
    event-keep:
      value-names:
        - '^/COUNTERS/[^/]+/(SAI_PORT_STAT_IF_IN_OCTETS|SAI_PORT_STAT_IF_OUT_OCTETS)$'
```

The change should include behavior tests, configuration decoding coverage, documentation, and a wide-event benchmark. Application-specific allow-lists remain outside gNMIc.

pood1e:
## What changed

- add an `event-keep` processor that retains matching value/tag names or values
  and removes unmatched entries in place
- leave values or tags untouched when that category has no configured selectors
- discard events left without values, tags, or delete paths while preserving
  delete-only events
- define `values` matching as string-only; non-string fields can be selected by
  `value-names`
- register and document the processor
- cover no-op, name matching, string value matching, OR semantics, independent
  tag/value filtering, empty and delete-only events, nil events, invalid regexes,
  config loading, and numeric value behavior
- add a 34,014-field benchmark matching the production-shaped wide-event case

## Why

`event-allow` filters whole events and `event-delete` requires enumerating every
unwanted field. Wide telemetry events need a native allow-list before
Starlark/JQ transformations so their cost scales with exported fields instead
of the full vendor payload.

Benchmark on the submitted implementation:

```text
BenchmarkKeepWideEventValueNames-16  385  2941781 ns/op  2624901 B/op  131 allocs/op
BenchmarkKeepWideEventValueNames-16  416  2783426 ns/op  2624813 B/op  130 allocs/op
BenchmarkKeepWideEventValueNames-16  435  2715901 ns/op  2624811 B/op  130 allocs/op
```

The benchmark includes cloning the 34,014-entry input map on every iteration;
the processor then reduces it to two retained fields.

## Validation

- `go test -race ./pkg/formatters/... ./pkg/config`
- `git diff --check`

Closes #950.

### DIFF
--- a/docs/user_guide/event_processors/event_keep.md
+++ b/docs/user_guide/event_processors/event_keep.md
@@ -2,6 +2,9 @@ The `event-keep` processor removes tags or values that do not match the configur
 
 Selectors for names and values use OR semantics. Tags are filtered only when `tag-names` or `tags` is configured; values are filtered only when `value-names` or `values` is configured. An unconfigured category is left unchanged.
 
+`values` regular expressions match string values only. Retain numeric, boolean, structured,
+or other non-string values with `value-names`.
+
 ```yaml
 processors:
   keep-interface-counters:
--- a/docs/user_guide/event_processors/event_keep.md
+++ b/docs/user_guide/event_processors/event_keep.md
@@ -1,0 +1,32 @@
+The `event-keep` processor removes tags or values that do not match the configured regular expressions. It complements `event-delete` and is useful when an event contains many fields but only a small allow-list is needed.
+
+Selectors for names and values use OR semantics. Tags are filtered only when `tag-names` or `tags` is configured; values are filtered only when `value-names` or `values` is configured. An unconfigured category is left unchanged.
+
+```yaml
+processors:
+  keep-interface-counters:
+    event-keep:
+      value-names:
+        - '^/interfaces/[^/]+/(in-octets|out-octets)$'
+      tag-names:
+        - '^(resource_id|source)$'
+```
+
+Given the following event:
+
+```json
+{
+  "tags": {
+    "resource_id": "leaf-1",
+    "source": "192.0.2.1:57400",
+    "subscription-name": "interfaces"
+  },
+  "values": {
+    "/interfaces/ethernet-1/in-octets": 42,
+    "/interfaces/ethernet-1/out-octets": 24,
+    "/vendor/debug": "ignored"
+  }
+}
+```
+
+the processor keeps the two interface counters and the `resource_id` and `source` tags.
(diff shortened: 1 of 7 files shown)


karimra:
Thanks for the contribution. 
One question: after filtering, an event can be left with zero values and zero tags and is still emitted. Shouldn't it be entirely filtered out in that case?

pood1e:
Addressed in `3ed66a96a8013cb98c5d7180ac5fc4f12a0317c4`.

`event-keep` now discards events that become truly empty after filtering. Delete-only events remain valid, and an empty selector configuration remains a no-op. Regression coverage verifies empty-event removal, delete preservation, and ordered batch compaction.

### Test validation

- Head commit: `3ed66a96a8013cb98c5d7180ac5fc4f12a0317c4`
- `go test -race ./pkg/formatters/... ./pkg/config` — PASS
- `git diff --check` — PASS
- GitHub checks: [`check-changes`](https://github.com/openconfig/gnmic/actions/runs/33849610875/job/100949166949) and Google CLA — PASS

karimra (approved):

PR #952, opened 2026-08-28
perf(processor): avoid condition JSON round trip

### LINKED ISSUE #949: Processor conditions JSON-round-trip entire events and cause excessive allocations
## Bug description

`formatters.CheckCondition` serializes every `EventMsg` to JSON and immediately deserializes it into a map before evaluating a gojq condition. The cost is paid for every conditional processor and scales with the complete values payload even when the condition only reads `.name` or `.tags`.

A benchmark with one event containing 34,014 values and the condition `.name == "port-counters"` on current `main` (`c820a545`) reports approximately:

```text
28-31 ms/op
13-14 MB/op
170,380 allocations/op
```

Evaluating against the existing `EventMsg.ToMap()` representation instead reports approximately:

```text
0.67-1.03 ms/op
1.8 KB/op
18 allocations/op
```

## Type

Performance bug.

## Impact

Large telemetry notifications can saturate collector CPU and create substantial GC pressure before the actual processor runs. Pipelines that use `event-allow` and `event-combine` multiply the cost for the same event. This can prevent collectors from keeping up even when downstream output queues apply backpressure.

## Proposed fix

Run the compiled gojq condition against `EventMsg.ToMap()`, which is already the event representation used by `event-jq`, and add field-level regression coverage plus a large-event benchmark.

pood1e:
## What changed

- remove the full-event JSON marshal/unmarshal round trip from every processor
  condition evaluation
- evaluate compiled conditions against the existing `EventMsg.ToMap()` schema
- clone mutable containers before passing them to `gojq`
- preserve JSON byte-slice semantics by exposing `[]byte` values as Base64
  strings, including nested values
- normalize other non-native gojq values with a localized JSON fallback
- add regression coverage for JSON container semantics, byte values, rejected
  unsupported values, and input immutability
- add a 34,014-value benchmark

## Why

Conditions commonly inspect only `.name` or `.tags`, but the previous path
serialized and decoded the complete event for every conditional processor.
Large telemetry notifications therefore incurred full-payload allocation and GC
cost multiple times before the selected processor ran.

Passing the event directly is unsafe because `gojq` normalizes numbers in maps
and slices in place. Passing `EventMsg.ToMap()` without normalization also
changes the established representation of byte slices from Base64 strings to
raw `[]byte` values. The submitted implementation clones mutable containers,
converts byte slices explicitly, and uses JSON conversion only for uncommon
non-native values.

Production-shaped benchmark on current `main` before the change:

```text
28-31 ms/op
13-14 MB/op
170,380 allocs/op
```

After the safe conversion:

```text
BenchmarkCheckConditionLargeEvent-16  1372  2475615 ns/op  2626553 B/op  152 allocs/op
BenchmarkCheckConditionLargeEvent-16  1489  2474924 ns/op  2626557 B/op  152 allocs/op
BenchmarkCheckConditionLargeEvent-16  1513  2642596 ns/op  2626553 B/op  152 allocs/op
```

## Validation

- `go test -race ./pkg/formatters/...`
- `go test ./pkg/formatters -run '^$' -bench '^BenchmarkCheckConditionLargeEvent$' -benchmem -benchtime=3s -count=3`
- `git diff --check`

Closes #949.

### DIFF
--- a/pkg/formatters/condition.go
+++ b/pkg/formatters/condition.go
@@ -1,0 +1,136 @@
+// © 2026 Nokia.
+//
+// This code is a Contribution to the gNMIc project (“Work”) made under the Google Software Grant and Corporate Contributor License Agreement (“CLA”) and governed by the Apache License 2.0.
+// No other rights or licenses in or to any of Nokia’s intellectual property are granted for any other purpose.
+// This code is provided on an “as is” basis without any warranties of any kind.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package formatters
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"math"
+
+	"github.com/itchyny/gojq"
+)
+
+func CheckCondition(code *gojq.Code, event *EventMsg) (bool, error) {
+	if code == nil {
+		return true, nil
+	}
+
+	input, err := conditionInput(event)
+	if err != nil {
+		return false, err
+	}
+	result, ok := code.Run(input).Next()
+	if !ok {
+		return false, nil
+	}
+	if err, ok := result.(error); ok {
+		return false, err
+	}
+	matched, ok := result.(bool)
+	if !ok {
+		return false, fmt.Errorf("unexpected condition return type: %T | %v", result, result)
+	}
+	return matched, nil
+}
+
+func conditionInput(event *EventMsg) (map[string]interface{}, error) {
+	input := event.ToMap()
+	if input == nil {
+		return nil, nil
+	}
+	cloned, err := cloneConditionValue(input)
+	if err != nil {
+		return nil, err
+	}
+	return cloned.(map[string]interface{}), nil
+}
+
+// cloneConditionValue copies mutable containers and converts values that are
+// not native gojq inputs while preserving encoding/json byte-slice semantics.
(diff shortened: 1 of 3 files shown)


karimra:
There is a regression here. JSON unmarshal turns []bye values in a base64 string, ToMap() does not.

pood1e:
Addressed in `4476fddde787128e6870e98e51db103ee42c29e1`.

Condition inputs now retain the established JSON representation of byte slices without restoring the full-event JSON round trip. Top-level and nested `[]byte` values become Base64 strings, mutable containers are cloned before `gojq` normalization, and uncommon non-native values use a localized JSON fallback. Regression coverage also verifies rejected unsupported values and input immutability.

### Test validation

- Head commit: `4476fddde787128e6870e98e51db103ee42c29e1`
- `go test -race ./pkg/formatters/...` — PASS
- `git diff --check` — PASS
- `go test ./pkg/formatters -run '^$' -bench '^BenchmarkCheckConditionLargeEvent$' -benchmem -benchtime=3s -count=3` — PASS; 2.48–2.64 ms/op, 2,626,553–2,626,557 B/op, 152 allocs/op
- GitHub checks: [`check-changes`](https://github.com/openconfig/gnmic/actions/runs/33849608005/job/100949158019) and Google CLA — PASS

karimra (approved):

PR #954, opened 2026-08-28
feat(prometheus-write): add stream-aware sampling before conversion

### LINKED ISSUE #953: Prometheus write output converts wide snapshots before size-aware sampling
### Bug description

The `prometheus_write` output receives `proto.Message` values through `Write`, queues them, converts each response into events, and only then runs event processors. A processor can rate-limit or discard an event, but it cannot avoid the allocation and GC cost of converting a wide snapshot that will be discarded.

This is severe for telemetry servers that emit large counter-table notifications at a higher rate than the required metric cadence. One observed SONiC COUNTERS notification flattened to about 34,000 values. Some servers then emit narrow incremental notifications on the same subscription, so sampling every message by subscription would incorrectly drop required deltas.

Type: performance / design

### Impact

- CPU and allocation work scale with wide-notification rate instead of the required metric cadence.
- Output queues and gRPC flow control can lag, making otherwise healthy counter metrics stale.
- Event processors cannot solve the conversion cost because they run afterward.
- Unconditional subscription-level sampling is unsafe for streams that mix snapshots and deltas.

### Proposed solution

Add an opt-in `message-sampling` policy to the Prometheus write output:

- configure an interval and optional minimum protobuf size by subscription name;
- key state by source and subscription;
- make the decision in `Write` before the protobuf message enters the conversion queue;
- let messages below the configured size bypass sampling without consuming the interval;
- optionally assign each key a stable phase to avoid synchronized fleet bursts;
- bound state with an LRU cache;
- roll back an accepted timestamp if enqueue fails;
- expose a skipped-message counter;
- leave behavior unchanged when the option is absent.

A zero size threshold samples all matching messages and is only suitable when every notification is self-contained. A nonzero threshold supports mixed streams by sampling wide snapshots while preserving narrow deltas.

A linked PR includes unit/race tests and documentation.

pood1e:
## What changed

- add opt-in `message-sampling.by-subscription` rules to the Prometheus write
  output
- configure each rule with an `interval` and optional `minimum-bytes` threshold
- sample managed gNMI STREAM Update responses by source and subscription before
  protobuf-to-event conversion
- preserve every Update response before the first successful SyncResponse
- treat each reconnect as a new stream instance with a new initial snapshot
- let narrow messages, control responses, ONCE/POLL subscriptions, and unmanaged
  messages bypass sampling without consuming the interval
- align later matching messages to stable per-stream phases when `spread` is
  enabled
- bound sampler state with an LRU cache and roll back accepted state when
  enqueueing is canceled
- retain sampler state across unrelated output configuration updates
- expose `gnmic_prometheus_write_output_messages_skipped_total`

## Why

Event processors run after protobuf conversion. High-frequency wide snapshots
therefore consume CPU and allocations even when a later processor discards them.
Some telemetry streams mix a multi-response initial snapshot with narrow
incremental notifications, so sampling must preserve the complete initial sync
and must not consume cadence for narrow deltas.

After initial sync, every matching response at or above `minimum-bytes` must be
independently discardable. Sampling is not safe for a later logical snapshot
that is split across multiple responses because the output cannot infer that
grouping.

## Configuration

```yaml
message-sampling:
  by-subscription:
    port-counters:
      interval: 60s
      minimum-bytes: 262144
  cache-size: 100000
  spread: true
```

## Validation

- `go test -race ./...`
- `cd pkg/api && go test -race ./target`
- `git diff --check`

Fixes #953.

### DIFF
--- a/docs/user_guide/outputs/prometheus_write_output.md
+++ b/docs/user_guide/outputs/prometheus_write_output.md
@@ -95,9 +95,42 @@ outputs:
     num-workers: 1
     # an integer, sets the number of writers draining the buffer and writing to Prometheus
     num-writers: 1
+    # optional source/subscription sampling performed before event conversion.
+    # A size floor allows narrow incremental updates to bypass sampling.
+    message-sampling:
+      by-subscription:
+        port-counters:
+          interval: 15s
+          minimum-bytes: 262144
+      cache-size: 100000
+      spread: true
 ```
 
 `gnmic` creates the prometheus metric name and its labels from the subscription name, the gnmic path and the value name.
+
+## Message Sampling
+
+`message-sampling.by-subscription` maps subscription names to an interval and an optional
+protobuf message size floor. Matching messages at or above `minimum-bytes` are sampled per
+source and subscription before event conversion and processing. Smaller messages bypass
+sampling and do not consume the sampling window, which preserves narrow incremental updates
+from streams that also emit wide snapshots. A zero `minimum-bytes` samples every matching
+message.
+
+Choose the size floor from observed message sizes and keep it below the smallest wide
+snapshot that must be sampled. Sampling all messages is appropriate only when every message
+is independently complete; partial update streams require a non-zero size floor or must not
+use this option.
+`cache-size` bounds the number of source/subscription timestamps retained in memory and
+defaults to `100000`. Skipped messages are reported by
+`gnmic_prometheus_write_output_messages_skipped_total`.
+
+`spread: true` assigns each source/subscription stream a stable phase inside its interval.
+The first matching wide message is always accepted because it may be the stream's only
+complete baseline. Later wide messages align to the stable phase, avoiding synchronized
+conversion and write bursts. Enforcing the interval while moving to that phase can delay
+the second accepted wide message by less than two intervals; narrow messages still bypass
+sampling throughout that period.
 
 ## Metric Generation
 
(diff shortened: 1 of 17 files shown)


pood1e:
Follow-up hardening is in `90775bb133cc954c8b9d58507ab02bdae2aed463`:

- use structured, trusted source/subscription identity instead of mutable event metadata;
- isolate sampling state by source, subscription, and stream attempt so delayed responses from reconnects cannot reset the active window;
- reserve sampling admission and commit it only after the output queue accepts the message, so rollback cannot evict unrelated LRU state;
- reuse one subscription context across outputs and remove the hot-path concatenated-key allocation;
- add regressions for interleaved reconnects, transactional rollback at LRU capacity, output-manager propagation, queue timeout rollback, and metadata overrides.

Validation completed on this head:

- `go test -race ./...`
- `(cd pkg/api && go test -race ./...)`
- `(cd pkg/cache && go test -race ./...)`
- `go vet ./...` in the root, `pkg/api`, and `pkg/cache` modules
- `go test -race -count=100 -run 'TestSubscriptionSampler|TestPromWriteOutputSampling' ./pkg/outputs/prometheus_output/prometheus_write_output`
- synthetic merge with current `origin/main` (`fa4eb451b28ed15f6d6e69a281f8cbbe5a781073`), followed by full race tests in all three modules

The rejected-message benchmark reports `0 B/op` and `0 allocs/op` on the sampling hot path.

karimra:
While I get the problem you are trying to solve, I have to say I'm not a fan of the solution.
The sampler filters by source and subscription and doesn't consider wildcard keys that might be present in the subscription. If responses get sampled, it can result in entire keys being sampled out.

I think the solution you are looking for is running 2 gNMIc instances where the first one subscribes to the target at a fast sample interval (say 10s) while the second one subscribes the first gNMIc instance at the desired sampling rate (say 60s). 
The second instance writes to prometheus and the first instance write to whichever other output is fine with frequent writes.
If you don't need a frequent subscription, you might as well just subscribe with a 60s sample interval from the target.

Issue #240, opened 2023-10-05
Silent drop of optical power Events with "-inf" value from Juniper MPC7 line cards.


flyboyjon:
Version 0.32.0

When subscribing to a Juniper MX router using path **"/junos/system/linecard/optics/"** to obtain transceiver optical information, the interfaces on this MPC7E line card report the metric **"lane_laser_receiver_power_dbm"** with a value **"-inf"**.  This is an unexpected value, normally this metric would contain a float value of 0 or possibly -40 when no light is received. That is the observed behaviour on different model of line cards in this same chassis.

The event is silently dropped (no error) when using **file** output with `--format=event`, when using `--format prototext` the metric is visible in the file.

The only way we discovered this is that the Prometheus output reports this error in the log file :  

`2023/10/05 22:06:36.648677 [prometheus_output:prometheus] failed to convert message to event: invalid character 'i' in numeric literal`

There does not seem to be a similar validation error for the file output, which is why we didn't notice it.

I don't seem to be able to apply a Processor to normalise this value, so the whole metric is missing from output when we use `--format=event` in order to perform event filtering and processing.

<img width="1506" alt="update_message" src="https://github.com/openconfig/gnmic/assets/73797573/35c72217-f9ab-4f28-ba90-6dc2211a2f11">

So, a couple of observations:
1. Metric was silently dropped with no error in the log file when using the file output
2. Metric was dropped with an error in the logfile when using Prometheus or Influx outputs
3. It was impossible to identify the offending metric in the incoming data without some extra print statements in prometheus_output.go:304 to dump the raw message and identify the value causing the issue.

Not sure what the fix could be here, since the device is not reporting a numerical value on this metric as it does on other line card/interfaces, but on the other hand, Juniper is a common enough platform that I suspect others will run into this issue too.  At least the error needs to be logged somewhere for file outputs that use --format=event so that we know the metric was dropped.

karimra:
The issue here is that `-inf` is not a valid json value, it has to be quoted `"-inf"`. The standard go json lib cannot unmarshal it.

flyboyjon:
Agreed, the device is sending bad data. 

As far as I can see though, I cannot capture this and correct it in a Processor, because gNMIC is dropping the Event before it gets to the Processors ?  Please confirm my understanding that Processors work only on Event formats. Thanks.

karimra:
That's correct. Can the device send the same data with a different encoding?

flyboyjon:
[...]

PR #963, opened 2026-09-03
fix(influxdb_output): drain client errors off the write goroutine

peejaychilds:
## Problem

Two or more unencodable points in quick succession permanently stop an influxdb output writing, for every target on the collector, while `gnmic_target_up` still reports 1 and no further log output appears.

The client reports point-encoding failures with an unconditional send on a single-slot error channel (influxdb-client-go v2.14.0, `api/write.go:254-263`, `errCh: make(chan error, 1)`). `worker()` drained that channel from the same
`select` loop that calls `WritePoint`. 

The first unencodable point parks its error in the slot; the second blocks `WritePoint` on the send, and the goroutine stuck there is the only one that could drain the error it is waiting to deliver. 

`eventChan` is unbuffered, so `Write()` then backpressures, and `App.export()` holds `operLock.RLock()` across that call, which stalls every target while the gRPC connections stay `READY`.

A single point recovers, since it fits the slot and is drained on the next iteration.

It takes two or more before the first is collected, which a burst of dark optical lanes in one sample provides: a dark lane reports `10*log10(0)` = `-Inf` for received power.

## Changes

- drain the client error channel on a dedicated goroutine, joined by the worker so `Close()` still waits for it. Any encode error can wedge the worker, not only non-finite floats
- drop `NaN`/`±Inf` float fields before the point is built, keeping the other fields on that point so one dark lane does not discard a whole sample
- count drops in `gnmic_influxdb_output_non_finite_values_dropped_total`, and log once then summarise every 5 minutes rather than per event
- add `Org`/`Bucket` to `clientNeedsRebuild`, and bind one WriteAPI per worker generation: each org/bucket pair gets its own WriteAPI with its own error channel 
- call `registerMetrics`/`initMetricLabels` from `Update`, so enabling `enable-metrics` by config reload exposes the counter
- document `enable-metrics` for this output, previously marked `NOT IMPLEMENTED`

`kafka_output` already drains its sarama `Errors()` channel on a separate goroutine, so this makes influxdb consistent with it.

## Validation

- `go test -race -count=2 ./pkg/outputs/influxdb_output/` — 18 tests
- `go vet ./pkg/outputs/influxdb_output/`
- each new test re-run with the fix disabled, confirming it fails
- run against an MX960 with a dark lane, no event processors, writing through telegraf   into InfluxDB: field dropped, siblings written, counter incrementing, receive counter climbing, 48 drops logged as 5 lines over 21 minutes

[`#240`](https://github.com/openconfig/gnmic/issues/240) is the same value but a different failure — silently dropped under `json` encoding, where under `proto` it reaches the output and wedges it.

### DIFF
--- a/pkg/outputs/influxdb_output/influxdb_output.go
+++ b/pkg/outputs/influxdb_output/influxdb_output.go
@@ -28,6 +28,7 @@ import (
 
 	influxdb2 "github.com/influxdata/influxdb-client-go/v2"
 	"github.com/openconfig/gnmi/proto/gnmi"
+	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/openconfig/gnmic/pkg/api/types"
 	"github.com/openconfig/gnmic/pkg/api/utils"
@@ -91,6 +92,9 @@ type influxDBOutput struct {
 
 	store        store.Store[any]
 	healthCancel context.CancelFunc
+	reg          *prometheus.Registry
+
+	nonFiniteLog nonFiniteLogState
 }
 
 func (i *influxDBOutput) init() {
@@ -183,6 +187,7 @@ func (i *influxDBOutput) Init(ctx context.Context, name string, cfg map[string]i
 	}
 
 	i.store = options.Store
+	i.reg = options.Registry
 
 	if newCfg.Name == "" {
 		newCfg.Name = name
@@ -200,6 +205,11 @@ func (i *influxDBOutput) Init(ctx context.Context, name string, cfg map[string]i
 
 	// store config
 	i.cfg.Store(newCfg)
+
+	if err := i.registerMetrics(); err != nil {
+		return err
+	}
+	i.initMetricLabels()
 
 	// build dynamic config
 	dc := new(dynConfig)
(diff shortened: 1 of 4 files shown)


pkg/outputs/influxdb_output/influxdb_output.go
```diff
@@ -770,6 +787,22 @@ START:
...
+	// clear of WritePoint, or a pending error strands the worker there.
+	stopErrDrain := make(chan struct{})
+	drainDone := make(chan struct{})
+	go i.drainWriteErrors(stopErrDrain, errCh, idx, drainDone)
+	var stopOnce sync.Once
+	stopDrain := func() {
+		stopOnce.Do(func() { close(stopErrDrain) })
+		<-drainDone
+	}
+	defer stopDrain()
```
karimra, line 804: On reset, the loop goes to START and this defer is never called. So N resets will stack N drain loops that only stop when the worker exists.

karimra (changes requested):

karimra (approved):

karimra:
Thanks for the fix!

PR #980, opened 2026-09-06
Fix Kubernetes locker discovery with EndpointSlices

pood1e:
Kubernetes locker discovery uses the deprecated core/v1 Endpoints API. This changes direct lookup and peer watching to discovery.k8s.io/v1 EndpointSlices selected by Service name.

The client-go informer maintains the slice set across watch reconnects and expired resource versions. Discovery publishes deterministic, deduplicated snapshots, including empty results after deletion; it filters endpoint conditions and formats IPv6 addresses correctly. The guide documents the Service contract and namespaced RBAC. Lease acquisition, renewal and election logic are unchanged.

Local validation of implementation `92a5d658fa2cdb4b418e5e2ab0bdbaf7e67bdd0b`:

- `./tests/run_tests.sh`
- `go test -race ./pkg/lockers/k8s_locker ./pkg/app`
- `go vet ./pkg/lockers/k8s_locker`
- Pod-replacement and peer-recovery checks with EndpointSlice permissions and no core/v1 Endpoints permission.

Regression tests cover slice aggregation, Service/namespace selection, endpoint conditions, deduplication, add/update/delete, empty discovery, watch closure, resource-version expiry and cancellation. The documentation-only follow-up at `9ea52f6610a5083c3dbd485fd80b0c848f3c90c6` has parsed YAML examples. GitHub Actions requiring maintainer approval remain pending.

Fixes #962.

PR #983, opened 2026-09-06
Fix Kubernetes locker ownership with client-go Lease election

### LINKED ISSUE #982: Kubernetes locker cannot safely maintain target ownership
## Bug

The Kubernetes locker rejects valid application target identities and does not bound renewal requests by the remaining Lease validity window. Delayed coordination API requests can leave collection active after ownership has expired.

Type: functionality / correctness / scalability. Impact: a target may fail to acquire ownership or may continue collecting after its Lease expires. This issue consolidates #981.

## Reproduction: target identity

Run gNMIc with the `k8s` locker, its API server, an ordinary subscription named `probe`, and the documented Lease permissions. Submit this generic example:

```sh
curl -X POST http://localhost:7890/api/v1/config/targets \
  -H 'Content-Type: application/json' \
  -d '{"name":"Router-A","address":"192.0.2.1:57400","insecure":true,"subscriptions":["probe"]}'
curl -X POST http://localhost:7890/api/v1/targets/Router-A
```

The target API accepts the identity, but Lease creation fails because the derived Kubernetes resource name contains uppercase characters. The example address is reserved for documentation; the failure occurs before dialing. Use the runtime API for this case because Viper normalizes YAML map keys.

The encoding only replaces `/` with `-` and also uses the result as a label key. Other legal application identities can exceed Kubernetes name constraints or collide after replacement.

## Renewal failure mechanism

`KeepLock` makes synchronous API requests without a request deadline tied to Lease validity. Requests delayed by the client limiter or API server can block renewal and delay the ownership-loss notification. `Lock` also captures its acquisition timestamp before those requests complete. Query and renewal traffic share the same client request budget.

These paths are in [`Lock`, `KeepLock` and `Unlock`](https://github.com/openconfig/gnmic/blob/b796d09d5ebcc3e65faee82616e3f4ad5aebf933/pkg/lockers/k8s_locker/k8s_locker.go) and [`IsLocked`/`List`](https://github.com/openconfig/gnmic/blob/b796d09d5ebcc3e65faee82616e3f4ad5aebf933/pkg/lockers/k8s_locker/k8s_registration.go).

## Expected behavior

- Preserve application identities with bounded, collision-resistant Lease names and original-key metadata.
- Bound renewal requests and promptly cancel collection when ownership cannot be maintained.
- Release only the owned Lease, with Kubernetes concurrency preconditions.
- Use a deliberate API request budget and cached ownership queries.
- Cover identity boundaries, delayed/failed API requests, cancellation and ownership loss with regression tests.

#983 implements these changes. EndpointSlice discovery is handled separately by #980 / #962.

pood1e:
Fixes #982.

The Kubernetes locker can reject valid runtime target identities and can continue collection after Lease renewal is no longer reliable. This replaces the custom acquisition and renewal loop with client-go leader election and `LeaseLock`, gives every acquisition a unique holder identity, and reports renewal-deadline failures through the collector's stop-and-retry path.

Release verifies the session owner, uses UID and resource-version preconditions, and gives the elector shutdown, Lease read, and Lease delete independent timeout budgets. Shutdown releases run with bounded concurrency. The collector cancels its subscription before release I/O, outside the operational mutex.

A shared Lease informer serves ownership queries. Its API list/watch is restricted to `app=gnmic`, and an original-key prefix index avoids scanning every cached Lease for each `List` call. `qps` and `burst` make the client-go request budget explicit.

Existing keys supported by the previous locker keep their slash-to-hyphen Lease names and legacy labels, while annotations retain the exact key and value. Previously unsupported keys use `gnmic-<sha256>` names. This lets old and new replicas coordinate during a rolling update. The deprecated `renew-period` and `retry-timer` settings remain aliases during that rollout; the documented sequence removes them after every replica is upgraded.

Kubernetes setup, lifecycle constraints, RBAC, limitations, and the rolling-upgrade sequence are documented in `docs/user_guide/ha_kubernetes.md`, alongside the EndpointSlice discovery documentation merged in #980.

Local validation at `359cd3e1a5ff37834279a218e2f675791e576a3f`:

- `./tests/run_tests.sh`: passed for all three Go modules.
- `./tests/run_tests.sh --race`: passed for all three Go modules.
- `go vet ./pkg/lockers/k8s_locker ./pkg/app ./pkg/collector/managers/targets ./pkg/collector/managers/cluster`: passed.
- Staticcheck on the changed packages has no new findings; the unsuppressed run reports the repository's existing `pkg/app/app.go` SA1019 deprecation.
- Regression tests cover cancellation channel semantics, renewal failure, independent release budgets, bounded shutdown concurrency, rolling Lease compatibility, prefix-index filtering, ownership-safe deletion, and configuration aliases.

The earlier comments contain combined Kubernetes deployment evidence from this change and #980 before the review-fix rebase. The review fixes above are covered by the new local unit and race runs; GitHub checks on the updated head remain authoritative.


pood1e:
Local runtime checks passed for `2ba66e42ce82979aab28f3266add4122fb1687a4`, combined with the EndpointSlice implementation from #980 at `92a5d658fa2cdb4b418e5e2ab0bdbaf7e67bdd0b`:

- Target identities are preserved when mapped to Lease names.
- Pod replacement recovers active target ownership and fresh Subscribe responses.
- Failed Lease writes cause collection to stop; collection recovers after write access is restored.
- Sampled active owners agree with Lease annotations at converged checkpoints.

The stop check uses removal of active targets from the runtime API and unchanged receive counters, since UP gauges update asynchronously. Blocked API requests and release I/O are covered by the regression tests in this PR. These observations describe the exercised cases, not an exactly-once guarantee.

pood1e:
Additional local integration checks passed for target ownership reconciliation, subscription recovery and Remote Write with this PR combined with #980. The PR's local unit and race checks also passed.

GitHub Actions requiring maintainer approval remain pending; these results describe local validation.

pood1e:
Final combined deployment validation is now complete for this Lease lifecycle change together with #980:

- Kubernetes v1.36.2+k3s1, 3 nodes, 6 gNMIc replicas distributed 2/2/2.
- 641 target Leases and 4,487 subscriptions remained healthy through a 7,204-second steady-state run.
- Active owners remained unique; Remote Write failures, dropped messages, subscription failures, and container restarts had zero increase.
- Dedicated member/leader replacement recovery passed (11.458s and 25.639s business-sample recovery, both below the existing 60s limit).
- Lease write denial still stopped active collection and recovery restored all owners in the earlier fault test; the final run completed with no cleanup errors.

The PR is mergeable and all GitHub checks are green.

@karimra, could you review this together with #980 when available?

pkg/lockers/k8s_locker/k8s_lease_lifecycle.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+		select {
+		case <-ctx.Done():
+			session.cancel(ctx.Err())
+		case <-session.stopped:
+		}
+		if err := ctx.Err(); err != nil {
+			errs <- err
+			close(done)
+			return
```
karimra: on `ctx.Done()` this sends `err` on `errs` chan. `errs` should only carry renewal failures.

pkg/lockers/k8s_locker/k8s_lease_lifecycle.go
```diff
@@ -0,0 +1,187 @@
...
+
+func (k *k8sLocker) release(ctx context.Context, key string, session *leaseSession) error {
+	ctx, cancel := context.WithTimeout(ctx, k.Cfg.RetryPeriod)
+	defer cancel()
+	session.cancel(context.Canceled)
+	select {
+	case <-session.stopped:
+	case <-ctx.Done():
+		return ctx.Err()
+	}
```
karimra: this context can be mostly consumed by `session.stopped`. That leaves almost nothing to `client.Get` and `client.Delete` coming afterwards.

pkg/lockers/k8s_locker/k8s_lease_cache.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (k *k8sLocker) List(ctx context.Context, prefix string) (map[string]string, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+	leases, err := k.leases.List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
```
karimra: would be good to filter during listing rather than getting all leases and filtering them afterwards.

pkg/lockers/k8s_locker/k8s_lease_lifecycle.go
```diff
@@ -0,0 +1,187 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	var errs []error
+	for key, session := range sessions {
+		if err := k.release(ctx, key, session); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return errors.Join(errs...)
+}
```
karimra, line 209: This functions uses a shared `5s` timeout context and passes it to `k.release`, which applies a `RetryPeriod` (2s), for each session.
Either use a bounded worker pool to release the sessions and size the context timeout accordingly, or size the timeout based on `len(sessions)`.

pkg/lockers/k8s_locker/k8s_lease_metadata.go
```diff
@@ -0,0 +1,51 @@
...
+
+	coordinationv1 "k8s.io/api/coordination/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	coordinationclient "k8s.io/client-go/kubernetes/typed/coordination/v1"
+)
+
+func leaseName(key string) string {
+	digest := sha256.Sum256([]byte(key))
+	return "gnmic-" + hex.EncodeToString(digest[:])
+}
```
karimra, line 29: This change breaks rolling upgrades.

docs/user_guide/ha_kubernetes_locker.md
```diff
@@ -0,0 +1,73 @@
...
+
+The resource name is now `gnmic-` followed by the SHA-256 digest of the original
+key. This supports case-sensitive, colon-containing and long target identities
+without lossy slash replacement or Kubernetes label restrictions.
+
+All members sharing a cluster must stop before this upgrade. The previous and
+new encodings refer to different Lease objects, so a mixed-version rolling
+upgrade would create independent ownership domains. After all old members have
+stopped, apply the new configuration and RBAC, then start the upgraded members.
+Obsolete Lease objects can be removed after confirming their holders have stopped.
```
karimra: this is not great... we should provide an upgrade path

karimra (changes requested):

pood1e:
Review feedback is addressed in `359cd3e1a5ff37834279a218e2f675791e576a3f`, rebased onto current `main` (including #980).

Validation on the updated head:

- `./tests/run_tests.sh` — passed for `.`, `pkg/api`, and `pkg/cache`.
- `./tests/run_tests.sh --race` — passed for `.`, `pkg/api`, and `pkg/cache`.
- `go vet ./pkg/lockers/k8s_locker ./pkg/app ./pkg/collector/managers/targets ./pkg/collector/managers/cluster` — passed.
- `go mod tidy -diff` and `git diff --check` — clean.
- Staticcheck on `./pkg/lockers/k8s_locker ./pkg/app` passes after excluding the existing `pkg/app/app.go:29` SA1019 deprecation; no finding is introduced by this PR.

The new regression coverage exercises caller cancellation without an `errs` send, renewal failure reporting, independent release timeout budgets, the 16-worker shutdown bound, legacy Lease interoperability during a rolling update, indexed prefix lookup, and deprecated configuration aliases.

The Test workflow is waiting for maintainer approval because this is a forked PR. I have left the review threads open for maintainer verification.

@karimra, all six review points have corresponding fixes and replies; this is ready for re-review when available.

pkg/lockers/k8s_locker/k8s_lease_lifecycle.go
pood1e: Fixed in `359cd3e1`. Caller cancellation now cancels the Lease session and closes `done` without sending on `errs`; `errs` is used only when the lock cannot be maintained. The blocking app-side error-channel drain was removed as well. `TestKeepLockCancellationCompletesWithoutErrorReceiver` covers this path.

pkg/lockers/k8s_locker/k8s_lease_cache.go
pood1e: Fixed in `359cd3e1`. The informer list/watch remains server-side filtered to `app=gnmic`, and now maintains an `original-key-prefix` index. `List` uses `ByIndex(prefix)`, so each call receives only matching cached Leases instead of enumerating the whole informer cache. `TestLeaseCacheFiltersAndTracksOwnership` checks the index cardinality and API-free reads.

pkg/lockers/k8s_locker/k8s_lease_lifecycle.go
pood1e: Fixed in `359cd3e1`. `release` now creates a fresh `RetryPeriod` timeout for each stage: waiting for the elector to stop, reading the Lease, and deleting it. `TestReleaseUsesIndependentTimeouts` verifies that the stop wait cannot consume the Get or Delete budget.

pkg/lockers/k8s_locker/k8s_lease_metadata.go
pood1e: Fixed in `359cd3e1`. Keys representable by the previous locker keep the same slash-to-hyphen Lease name and legacy label, while annotations carry the exact key/value. Digest names are used only for keys the previous implementation could not represent. The compatibility test starts from an old-format active Lease, verifies that a new replica cannot acquire a second object, and verifies dual metadata on newly created compatible Leases.

pkg/lockers/k8s_locker/k8s_lease_lifecycle.go
pood1e: Fixed in `359cd3e1`. `Stop` uses `errgroup.SetLimit(16)` to bound concurrent releases. Every release has independent bounded stop/Get/Delete stages, so one shared deadline no longer starves later sessions. `TestStopBoundsConcurrentReleases` verifies the worker limit and that all sessions complete.

docs/user_guide/ha_kubernetes_locker.md
pood1e: Reworked in `359cd3e1`. The stop-all instruction is gone. The existing `ha_kubernetes.md` now gives an ordered rolling path: grant watch RBAC first, retain supported keys and legacy timing fields while rolling the binary, then rename the deprecated fields after every replica is upgraded. It also documents the boundary for previously unsupported keys and the retained legacy collision behavior.

PR #1007, opened 2026-09-14
fix(clustering): preserve pending target assignments

### LINKED ISSUE #1006: [Bug] Target assignment retries can livelock Kubernetes Lease takeover
## Bug description

Cluster target assignment can livelock when acquiring an existing Kubernetes Lease takes longer than `target-assignment-timeout`.

`dispatchTarget` starts the target and polls the locker. When the assignment timeout is reached, it sends `DELETE /api/v1/targets/{name}` and immediately starts a new assignment. That cancellation is destructive for lockers whose acquisition is still legitimately pending.

This is deterministic with the client-go leader election implementation proposed in #983. For clock-skew safety, a new `LeaderElector` observes a foreign Lease for one complete lease duration before takeover. With a 15-second Lease and the default 10-second target assignment timeout, gNMIc cancels every elector before it can acquire the Lease. Each replacement creates a new elector and restarts the local observation window, so even a Lease whose `renewTime` is hours old is never acquired.

The HTTP loader and periodic dispatcher can also call `dispatchTarget` concurrently for the same target because only the periodic dispatcher uses `dispatchLock`. Their overlapping timeout loops cancel each other's target generations more frequently and the synchronous loader stops processing later targets.

## Type and impact

- Type: clustering correctness / concurrency / availability
- Severity: high
- Impact: after a collector restart, leader change, or target configuration change, one stale target Lease can remain unowned indefinitely and block later dynamic targets from being dispatched. Healthy devices then remain absent or keep stale runtime configuration.

## Reproduction

1. Run one clustered gNMIc instance with a dynamic loader.
2. Leave a target Lease from a previous process generation in Kubernetes.
3. Configure the Lease duration to 15 seconds and `target-assignment-timeout` to 10 seconds.
4. Use client-go leader election for target ownership as in #983.
5. Add or update the target through the loader.

Observed loop:

```text
POST /api/v1/targets/device-a -> 200
Attempting to acquire leader lease
max retries reached, reselecting
DELETE /api/v1/targets/device-a -> 200
failed to lock target: context canceled
```

The loop repeats every few seconds. The Lease resource version, holder identity, and renew time remain unchanged.

A deterministic regression can use a fake locker whose acquisition completes after the assignment observation timeout and HTTP handlers that count target starts and deletes.

## Expected behavior

- An accepted target whose lock acquisition is pending is not repeatedly deleted and restarted merely because the leader's observation timeout elapsed.
- Concurrent dispatch requests for the same target converge on one in-flight assignment.
- A slow target does not prevent the loader from dispatching unrelated targets.
- Explicit target deletion, service loss, and application shutdown still cancel acquisition and release ownership.

## Proposed direction

Separate assignment observation from target acquisition lifetime. Coalesce dispatch for the same target, keep an accepted acquisition alive across an observation timeout, and let later reconciliation observe its eventual owner. Bound loader concurrency so one slow target cannot block the rest of a snapshot.

This was found while dogfooding #983; I can submit the regression and fix.

pood1e:
## Summary

- Treat `target-assignment-timeout` as a bound on leader-side lock observation.
- Preserve an assignment already accepted by a cluster member when Lease ownership is not yet visible.
- Add a regression test proving the timeout returns without sending `DELETE` or restarting the assignment.

## Problem

Kubernetes Lease takeover may legitimately take longer than the configured target assignment timeout. In particular, client-go leader election observes a foreign Lease locally for a full lease duration before attempting takeover so it does not depend on the remote `renewTime` clock.

The current dispatcher treats an observation timeout as an assignment failure: it deletes the target from the member and immediately retries. When the assignment timeout is shorter than the Lease duration, every retry cancels the member's leader election before takeover is possible. The target can therefore livelock forever, and a sequential loader can also stop progressing to later targets.

This change keeps the accepted member assignment running. The dispatcher still returns after the configured observation timeout, allowing callers to make progress, while explicit target removal, service loss, and shutdown retain their existing cancellation behavior.

## Validation

- `./tests/run_tests.sh`
- focused clustering regression tests, 10 repetitions
- focused clustering regression tests with `-race`, 5 repetitions
- `go vet ./pkg/app`
- `go build -trimpath .`

Fixes #1006


pood1e:
Validation evidence for `410a377126962f59ced230d70cca9c6eaa0006a3`:

- `./tests/run_tests.sh`: passed.
- Target dispatch regression tests, 10 repetitions: passed.
- Target dispatch regression tests with the race detector, 5 repetitions: passed.
- `go vet ./pkg/app`: passed.
- `go build -trimpath .`: passed.

The regression fails against the pre-fix dispatcher because it repeatedly restarts the assignment until the parent context expires; it passes with one start, zero deletes, and a bounded typed timeout after this change.

